PR #13, opened 2017-08-06
update DMXOPL

loki666:


Wohlstand:
Thanks!
Also, what do you think about my rework of gen_adldata utility? And because of too slow Nuked emulator I made caching to just don't re-generate same twise because of so slow emulator

loki666:
nice speed improvement indeed
didn't check the implementation, hoping that it doesn't fail

Wohlstand:
Also, I have moved banks list into INI file that doesn't requires recompilation of entire too to do that generation

PR #19, opened 2017-09-01
DMXOPL3 v2.2

loki666:
one more. this guy is restless


Wohlstand:
Include this, I also updating my SDL Mixer X where I using this library 😋

PR #33, opened 2017-11-22
Adl play rework

loki666:
rework `adl_play` and `struct Setup` to not need a backup_sample buffer


loki666:
but it was already like this before my changes...
adl_play is outputting a stereo 16bit PCM stream

which test did you made?

Wohlstand:
I ran it via my demo tool (you can build it by CMake by turning an "WITH_MIDIPLAY" option, requires SDL2).

Wohlstand:
Also, when I tried to fix a timing, I found two another issues:
* When nextTick is zero, tick() turns into infinite loop!
* After fix of infinite loop, time ticking is desynchronius [left - your branch with some my fixes, right - my current master](http://wohlsoft.ru/docs/_files_for_posts/Misc/AdlMidi-new-adlplay/MOSKVA%20(left=adlplay%20right=stable).flac)

loki666:
that's what I was trying to do, the performance shouldn't be an issue

Wohlstand:
Look what I mean:
the splited "chunk" rectangle means this chunk was built by two generation requests.
chunks-division
The load means a delay generated by emulator by one call. Even two calls, the delay was split and final delay is same. In result we got two calls to generate same chunk.
I'll try some at my side and I'll show that if it will work and it will don't break synchronization.

loki666:
calling multiple times the emulator for each out buffer is not an issue, in the end the emulator generate exactly the request samples to fill the buffer.
the load on the CPU is the same for every out buffer requested which is better.

Wohlstand:
Okay, I have tried to get rid of backup samples buffer and I have successfully kept synchronous tick processing: when dumping WAV, the resulted output 1:1 same in both before and after.
However, one thing I have found:

With backup samples we are always requested 1024-sample chunks (which are was stripped down per every loop).
Now, we are requesting chunks of different size. Therefore a result I have found while comparison of two dumps (are synchronously, but on few notes there are different)



I see that since we began to request different-size chunks, we got tiny begin offset of some notes even MIDI timing is going 1:1 with no difference.

loki666:
Not sure I understand all you are saying.

Anyway, the small notes offset is in fact more precise, than it was using fixed chunk length.
Since you were always requesting 512 frames of the emulator, It was entirely possible that a note On should have started in the middle of this chunk.
Splitting the chunks to match the Midi Events is in fact alot more precise.
There is note on event, but alots of events could happens like key pressure, channel volume.

512 frames at 44.1kHz is about 11ms, but MIDI ticks resolution can be high than this

Wohlstand:
Then, I need to change the ticking logic in total: Instead of ticking by time period, every "tick" call will tick all events in current row and will return delay to wait for next row. Use this time period to generate and then tick next row.

loki666:
yes that's how it should works.

calling Tick() should process all the midi events at the current tick, and return the time before the next midi event.
generate audio from emulator for that time and then call Tick() again

PR #46, opened 2018-04-08
fix the check which identifies cmake as a legacy version

jpcima:
This version comparison is wrong.
I am on cmake 3.10, which is not inferior to 3.2, although by numeric comparison it is.
I replace this by VERSION_LESS/VERSION_GREATER which according to manual are present in cmake since 3.0.2 at least.


Wohlstand:
Thanks! You right, my fault!

PR #50, opened 2018-04-09
version string fix

### LINKED ISSUE #47: The library's version string is incorrect
There is currently a problem executing the following code.

```    fprintf(stderr, "ADLMIDI version %s\n", adl_linkedLibraryVersion());```

The output is this. A problem needs to be fixed with the version macros.

```ADLMIDI version OPNMIDI_VERSION_MAJOR.OPNMIDI_VERSION_MINOR.OPNMIDI_VERSION_PATCHLEVEL```

jpcima:
#47 Thanks, but actually there remains a second problem with preprocessing.
It takes a second layer of macro to expand the string. This is good for gcc. It should be checked whether msvc accepts it as well. (If you look at Boost's [stringize](https://www.boost.org/doc/libs/1_66_0/boost/preprocessor/stringize.hpp) msvc has a special case)


Wohlstand:
Then I think is needed a some sort of unit test and AppVeyor build with MSVC target like I made with a case of AudioCodecs project, except one that AC has no any unit tests.

**Trivia:** I never liked MSVC myself as it has own weird cases are must take individual attention while CLand, Inte-CC and GCC / MinGW are in most of things fully compatible to each other where MSVC has lots of own clunky things include [the inconsistency ('friendly template' thing as example where I had a strong fight versus MSVC)](https://github.com/WohlSoft/PtrList/blob/master/vptrlist.h#L63-L68) and bugs.

PR #54, opened 2018-04-11
support for 32 bit mixing, and multiple sample formats

### LINKED ISSUE #49: adl_generate for int32_t output?
As I tried to put adlmidi and multi-chip on trial, with a few somewhat aggressive MIDI files, I noticed that the mixed output of sample type int16_t will eventually saturate.

Even if my program outputs to floating point, I don't have ability to process it through some kind of limiter, because it comes out of the library as already clipped.

Can there be a mixing routine which uses a bigger integer type?
(alternatively, as I believe that was already suggested, an access to the outputs of individual chips?)

jpcima:
⚠️ NEEDS TESTING ⚠️ 

This is for 32bit mixing, which introduces the formatted output APIs.
It will enable to supports two formats of PCM (8, 16 bit) and float (single, double), as well as output to interleaved/non-interleaved buffers, and output to larger integer types than format.

The internal mixing is upgraded from 16bit to 32bit, to prevent premature saturation.
The chip interface receives new output methods in 32bit in addition to the existing 16 bit ones.

fixes #49


Wohlstand:
Try to play this music [MUS_TITL.rmi.zip](https://github.com/Wohlstand/libADLMIDI/files/1899752/MUS_TITL.rmi.zip) on Bank-14 which originally played dirty because of clipping, how it will sound now?

jpcima:
Here I've captured one on a file I have which does a lot of clips.
You can see it, it is triggered at top left of screen, just a tiny bit above level.

untitled

jpcima:
I know the above is a shit capture but I'll do it ok after I complete writing test stuff
I opened a testing branch in adljack

jpcima:
I verified all things in the checklist

There is a [test case](https://github.com/jpcima/adljack/blob/32bitmix-testing/common.h#L15) which one can uncomment to test all scenarios.

Next
- [ ] to show clipping has been resolved
- [ ] port to OPNMIDI and edit emulator methods for 32bit

jpcima:
No clipping. Tested with all emulator cores.

Wohlstand:
Okay, lemme I'll check out that, and yeah, as bank-14 I meant turn the embedded bank with ID 14 where synth-strings are clippish even on real chip.

jpcima:
This PR is probably good for merging. As for OPNMIDI, I'm doing the second PR, I'm well advanced in it already so don't bother. The biggest work is in the chip interfaces. It will need the same testing but there is already adljack as testbed so it should be fast.

PR #60, opened 2018-04-13
propagate the include directory of ADLMIDI

jpcima:
Modify CMake so the user of *target_link_libraries* will automatically receive the include path


CMakeLists.txt
```diff
@@ -253,6 +254,7 @@ endif()
 if(libADLMIDI_SHARED)
     add_library(ADLMIDI_shared SHARED ${libADLMIDI_SOURCES})
     set_target_properties(ADLMIDI_shared PROPERTIES OUTPUT_NAME ADLMIDI)
+    target_include_directories(ADLMIDI_shared PUBLIC ${libADLMIDI_SOURCE_DIR}/include)
     target_include_directories(ADLMIDI_shared PUBLIC ${SDLMixerX_SOURCE_DIR}/include)
```
Wohlstand: `${SDLMixerX_SOURCE_DIR}/include` as in libOPNMIDI must not be here, as this line was added by mistake, so, it must be `libADLMIDI_SOURCE_DIR` as you have been added.

PR #66, opened 2018-04-16
flexible emulator selection API

jpcima:
Not being entirely happy with the state of emulator-related APIs, so I imagined an extension of it, and I submit it to your comments.

**Before** I would enumerate the emulators by creating a fake player, invoke `switchEmulator` for 0 to count until a -1 return. I would avoid relying on hardcoded values of `ADL_Emulator`. In the future, it's imaginable that Nuked 1.7 will be entirely superseded and removed.

**Now** This is an API proposal which could be very well used without any enum `ADL_Emulator`.
It has a query ability for emulator and their characteristics.

Enter the concept of **profile**. You can use a combination of profile flags to find a matching emulator. See adlmidi.h
Exactly one emulator should match **FASTEST**, idem for **MOST_ACCURATE**. Note these two flags don't necessarily imply **FAST** and **ACCURATE** respectively.

**Rationale**: I just want to present normal users who don't care about detail, a generic setting of speed/quality compromise choice.

### DIFF
--- a/src/chips/nuked_opl3_v174.cpp
+++ b/src/chips/nuked_opl3_v174.cpp
@@ -87,7 +87,9 @@ int NukedOPL3v174::generateAndMix32(int32_t *output, size_t frames)
     return (int)frames;
 }
 
+const char *const NukedOPL3v174::staticEmulatorName = "Nuked OPL3 (v 1.7.4)";
+
 const char *NukedOPL3v174::emulatorName()
 {
-    return "Nuked OPL3 (v 1.7.4)";
+    return staticEmulatorName;
 }
--- a/include/adlmidi.h
+++ b/include/adlmidi.h
@@ -162,9 +162,35 @@ enum ADL_Emulator
     ADLMIDI_EMU_end
 };
 
+/* Returns chip emulator */
+extern int adl_chipEmulator(struct ADL_MIDIPlayer *device);
+
 /* Switch the emulation core */
 extern int adl_switchEmulator(struct ADL_MIDIPlayer *device, int emulator);
 
+enum ADL_EmulatorProfileFlag
+{
+    ADLMIDI_EMUPROFILE_FAST          = 1 << 0,
+    ADLMIDI_EMUPROFILE_FASTEST       = 1 << 1,
+    ADLMIDI_EMUPROFILE_ACCURATE      = 1 << 2,
+    ADLMIDI_EMUPROFILE_MOST_ACCURATE = 1 << 3,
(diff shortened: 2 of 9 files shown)


src/chips/nuked_opl3_v174.cpp
```diff
@@ -87,7 +88,7 @@ int NukedOPL3v174::generateAndMix32(int32_t *output, size_t frames)
     return (int)frames;
 }
 
-const char *NukedOPL3v174::emulatorName()
+int NukedOPL3v174::emulatorId()
 {
-    return "Nuked OPL3 (v 1.7.4)";
+    return ADLMIDI_EMU_NUKED_174;
```
Wohlstand: It's BAD idea to put dependency of chipset from ADLMIDI as I told, I use same chipset in both ADLMIDI and OPL3 Bank Editor where is completely different base. You better duplicate the stuff into chipset too, to make it be independent. And yeah, Don't break `emulatorName()` as it is mainly using in OPL3 Bank Editor and is more convenient to call it rather request entry from separate array.

jpcima:
Yes this was a really crappy code. Shameful display :smile: 
I arranged it in a way you may like better.

The indexes of the characteristics table is made to match the enum values. If an emu is not built it, there will be a hole in the table and it represented by a `name` set to NULL. I modify the implementation to account for this detail.

Wohlstand:
I'll try to polish this tomorrow.

Wohlstand:
This is very outdated thing. It needs a different strategy to make this: without of any static references between things, as those are shared with OPL3BE and will not be compatible.

jpcima:
> This is very outdated thing.

I don't mind closing as I don't have usefulness for this right now.

I have nonetheless few comments:

- the linking problem was, as I remember, resolved, and makes the code good for OPL3BE use

- the capability matching deserves rethinking in this code.
for example disable nuked and there will be no longer a "most accurate" emulator match.

- there's an ABI consideration about using the ADLMIDI_EMU_* enum, because these numbers must never change.

Since the count of emulators is made a fixed number `ADLMIDI_EMU_end`, and a program is linked to a shared library of libADLMIDI, it doesn't know of new emulators which are added in the library after updates.

You can see in my programming with libADLMIDI, I don't use enum values, I try emulators by index until a failure and I collect the results.
A problem is: in the future you remove an outdated emulator from libADLMIDI, and then the Emulators enum has holes, and this method fails.
(And you don't have another choice that to create a hole, considering enum numbers being part of ABI)

Moreover: this method is only working having none of emulators disabled

Globally, I think there's only fragile methods to access into emulators and it needs a rework to enable flexible selections.

Issue #65, opened 2018-04-16
MIDI playback in hard RT context


jpcima:
In a softsynth plugin, MIDI playback can happen in the same thread as audio processing.
adlmidi_midiplay must be modified for hard-RT to avoid all kinds of malloc.

To give an example, `activenotes` is stored in a std::map of uint8_t.
In a std::map `malloc` is required for each insertion, ie. for every MIDI note on.
Midi notes being limited to 127, it would be a better idea to use flat array storage.

Please assign me the issue.

jpcima:
There are multiple strategies to solve the problem of note keeping.
(In all cases I believe the array might be in fact faster because of low element count and cache)

1. the most obvious may be to replace `map` with `vector` reserved with 128 cells. The positive is that I only have to iterate up to size, but removal will take O(n).
2. Just make it an array of 128 and iterate everything.

I guess I safely ignore the consideration of ordering of the note array? originally std::map is ordered by ascending note number.

Wohlstand:
Third strategy:
* Create 128 array which will index all notes.
* Create `vector` which will contain a chain of `playing indexes` to iterate them.

Ow! just now I have found that each `NoteInfo` contains another map `std::map<uint16_t /*adlchn*/, Phys> phys;` which refers using physical chip(s) channels (one two-op voice, two two-op voices or two pieces of four-op voice). So, I'll check out an ability to replace it with fixed-size array and the count value.

jpcima:
Yes I'm aware of this other pesky little map.
I have already made its acquaintance in original Bisqwit code.

jpcima:
I made the hard-realtime branch. This is a simple transformation of `activenotemap` into a fixed array.
It could be organized a bit better for sure.
Do you approve this?

https://github.com/jpcima/libADLMIDI/commit/21fd71af4394f88522630d9a900dc1d8d8ab8625

Wohlstand:
I have continued stuff, and I removed the "phys" map with replacing it with two-element arrays, that also need to test out for any possible issues with it. Also, I gonna to move MIDI Sequencer which loads and plays MIDI files in-home of library into separated file, then into separated class that will be possible to use in other libraries separately.

jpcima:
Nice! I will look at the code that you have done, and try it. Thanks.

I know there remains the `users` map, from `AdlChannel`.
I have not completely understood the full purpose of this map. Is it a history of the current and old notes on the channel?
[...]

PR #67, opened 2018-04-16
transform the note map into a fixed array

jpcima:
This is the code discussed in #65 made into a PR.
It is step 1 into making ADLMIDI hard-RT compliant.
As far as I can tell this code plays back without issues.

Second, I need to think how to handle the Phys map which is present in every NoteInfo.


Wohlstand:
Gonna to take this to polish...

Wohlstand:
Just for an info: a memory usage before:
default

and after your change:
default

I'll try to replace that damned map that is left alone for now.

PR #68, opened 2018-04-17
fix Android.mk

loki666:


Wohlstand:
Yeah, I forgot to update it when I made some reorganizing in the library, thanks! :fox_face:

PR #69, opened 2018-04-17
hard-RT experiment to eliminate the user map

jpcima:
:warning: DO NOT MERGE :warning: 

Work on issue #65.
It's a request for review on a removal of the user map, replaced by a linked list made of cells taken from static storage.

:warning: It does not take action in case of exhaustion of buffer, and will crash. I marked these cases with `#warning`.
It must be determined what action to take for these cases.

### DIFF
--- a/src/adlmidi_private.hpp
+++ b/src/adlmidi_private.hpp
@@ -816,7 +816,6 @@ public:
         {
             LocationData *prev, *next;
             Location loc;
-            uint8_t     note;
             bool sustained;
             char ____padding[7];
             MIDIchannel::NoteInfo::Phys ins;  // a copy of that in phys[]
--- a/src/adlmidi_private.hpp
+++ b/src/adlmidi_private.hpp
@@ -646,6 +646,8 @@ public:
                 {
                     return !operator==(oth);
                 }
+
+                // Phys() : chip_chan(0), insId(0), pseudo4op(false) {}
             };
             //! List of OPL3 channels it is currently occupying.
             Phys chip_channels[MaxNumPhysItemCount];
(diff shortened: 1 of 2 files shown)


src/adlmidi_private.hpp
```diff
@@ -749,36 +751,111 @@ class MIDIplay
...
+        enum { users_max = 128 };
+        LocationData *users, *users_free_cells;
+        LocationData users_cells[users_max];
+        unsigned users_size;
+
+        bool users_empty() const
+            { return !users; }
+        LocationData *users_find(Location loc)
+        {
+            for(LocationData *user = users; user; user = user->next)
```
Wohlstand: As first item may be empty, why not to find the begin first? Or lets to save begin pointer in separate place?

Wohlstand: As first item may be empty, why not to find the begin first? Or lets to save begin pointer in separate place?

jpcima:
Just quick comment: there's a crash with the code, I searched for a MIDI which reproduces the problem fast, and here's one. [01-toto_1978-child's_anthem.zip](https://github.com/Wohlstand/libADLMIDI/files/1921025/01-toto_1978-child.s_anthem.zip)

src/adlmidi_private.hpp
jpcima: None of the items is empty. The empty list is indicated by a NULL head.

Wohlstand: Okay then, I just got a think what will happen when you will kick a first item?

jpcima:
It crashes with AddressSanitizer detection within the first 15 seconds.
I have a backtrace which is not very helpful. (it must be these commits, it does not crash in master)

```
==10839==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000004 (pc 0x55dc7a1cf8f6 bp 0x000000000000 sp 0x7ffc11b79d90 T0)
==10839==The signal is caused by a READ memory access.
==10839==Hint: address points to the zero page.
    #0 0x55dc7a1cf8f5 in MIDIplay::NoteUpdate(unsigned short, MIDIplay::MIDIchannel::activenoteiterator, unsigned int, int) ../src/adlmidi_midiplay.cpp:1440
    #1 0x55dc7a1d2424 in MIDIplay::KillOrEvacuate(unsigned long, MIDIplay::AdlChannel::LocationData*, MIDIplay::MIDIchannel::activenoteiterator) ../src/adlmidi_midiplay.cpp:2308
    #2 0x55dc7a1d355b in MIDIplay::PrepareAdlChannelForNewNote(unsigned long, MIDIplay::MIDIchannel::NoteInfo::Phys const&) ../src/adlmidi_midiplay.cpp:2229
    #3 0x55dc7a1d92b9 in MIDIplay::realTime_NoteOn(unsigned char, unsigned char, unsigned char) ../src/adlmidi_midiplay.cpp:1185
    #4 0x55dc7a1e7c70 in MIDIplay::HandleEvent(unsigned long, MIDIplay::MidiEvent const&, int&) ../src/adlmidi_midiplay.cpp:2088
    #5 0x55dc7a1e9675 in MIDIplay::ProcessEventsNew(bool) ../src/adlmidi_midiplay.cpp:1681
    #6 0x55dc7a1ea41a in MIDIplay::Tick(double, double) ../src/adlmidi_midiplay.cpp:809
    #7 0x55dc7a1b8d09 in adl_playFormat ../src/adlmidi.cpp:840
    #8 0x55dc7a1ab7e6 in main ../utils/midiplay/adlmidiplay.cpp:513
    #9 0x7f748d4cef49 in __libc_start_main (/usr/lib/libc.so.6+0x20f49)
    #10 0x55dc7a1adca9 in _start (/home/jpc/Documents/Projects/libADLMIDI/build/adlmidiplay+0x12ca9)
```

Wohlstand:
I'm now at home, gotta to pull your code to me to debug it...

Wohlstand:
Yeah, the crash happen because of the null pointer:

Began at here:
default


Here crash wasn't happen as you wasn't set a hook function, this code is not touched:
default

And now.... KABOOM!
You just now dereferenced a null pointer!
default

jpcima:
I have realized this much but it's not the root cause just a symptom.
I've realized the problem is in the RT development which was merged to master.
I'm investigating more.

jpcima:
@Wohlstand I don't receive a notification when you edit, only on a new post. So I don't always see information you try to tell me.

Problem on **master**, not this branch in particular:
Try adlmidiplay with MIDI above and bank 72.

```
--- ACTIVE ---
======
71
======
--- INFO ---
======
USERS (2)
NTH (0)
LOC (13,68) (nil)
======
--- FOO ---
======
```

Below the line `if(i->second.kon_time_until_neglible <= 0l)`, I dump the information which leads to dereferencing a NULL NoteInfo. There is only the active note (71), AdlChannel has 2 users among which none is equal to the user `i` (channel 13 note 68).
Hence a NULL for `Ch[ i->first.MidCh ].activenotes_find(i->first.note)`.

I was about to compare with an older master branch with std::map code.

Wohlstand:
Anyway, as library now going to be used by @loki666, I gonna make the `stable` branch without of changes related to this flow as this is experimental one. When this will be stable, I'll merge everything.

jpcima:
Tried sdl2-hg-2.0.8.r32.ffd52bb02bcc-1.
No difference.

jpcima:
Good! now it must be examinated the action to take when there is exhaustion of capacity in the users list. It's the code I have marked with `#warning`.

https://github.com/jpcima/libADLMIDI/blob/545a0efcd8317ed174baf89ca987821224b8c940/src/adlmidi_midiplay.cpp

There are 4 instances.

1. ?
2. ?
3. Seems trivial. Just don't insert the user in the list. Cab't be sustained if it doesn't exist.
4. ?

jpcima:
Here's a possibility for maybe the final hard-rt update.
If you want I can rearrange this branch into a cleaner commit log.

I believe it's this comment of `UpdateArpeggio` which holds the answer about what `users` is, which I looked for all along.

```
    // If there is an adlib channel that has multiple notes
    // simulated on the same channel, arpeggio them.
```

case 1: just ignore. I'm not sure what is the purpose of doing this in `Upd_Patch`. it's the one I'm most uncertain about.
case 2: can't mark the note sustained. ignore.
case 3: no user info -> just bend the note as normal
case 4: update arpeggio: no room into the channel to accommodate the note? `continue` to the next.

Wohlstand:
Arpeggio is works for all chips channels overflow and all users are must be unlimited. If it is becoming limited, then autokill shortest users you see, however, users are per channel, so, is very small probably of overflow.

jpcima:
> Arpeggio is works for all chips channels overflow and all users are must be unlimited

The hardcoded limit is currently 128. To give an idea, with your short midi file which has an avalanche of notes, the maximum users at any time where 8 on 1 chip, if my memory is correct. In practice, it is unlikely an exhaustion of capacity will ever happen.
The max user count went down to 1 with, I believe, 5-6 chips possibly?

It does not matter if exhaustion would result in playback glitches. It matters if it results in crash.
Which I think is not the case, from my limited understanding of code.

Wohlstand:
No matter count of chips:
* MIDI channel holds list of notes are "on" in MIDI side
* Those notes are interpreting to be played on physical channels of any of chips
* MIDI notes are unlimited and may have count more than total count of channels of all chips
* When chip channels are all in use, MIDI does arpeggio between of running notes to keep more notes playing on less count of chip channels

jpcima:
> No matter count of chips:

I find it curious that you say this. In my few experiences the chip count had an impact.
I can try again if you want numbers.
As I interpret it, the arpeggio (aka `users`) is defined per Adlchannel (not Midi channel). The number of Adlchannels is multiple of the chip count (23 * NumCards).

---

With all things said, I consider personally the current code satisfying. I tested it with a lot of rt playback, without problem.
I wait for you to validate it, but you must surely be busy now.
If you know of something now you want me to change, just tell me.

Wohlstand:
I'm at job usually between 10~11 AM and 19~21 PM UTC+3. Yeah, all `users` are merging to process arpeggio, and arpeggio turns on when free chip channels (AdlChannel) are all in use. MIDI channel can play unlimited notes count, chip channel allows to play one note only.

Wohlstand:
As this thing is completed and merged through another pull-request, I'll close this.

PR #71, opened 2018-04-18
Phys code rewrite

jpcima:
I rewrite your commit https://github.com/Wohlstand/libADLMIDI/commit/09bb3bc4d31ac925a13f51f8fc4944330da158bb to try to stick to original map semantics of ADLMIDI.
It fixes crashes I have occuring in playback.

I call for you to verify this; you're probably the one person there is second to Joel to understand how all this stuff works.

(what a time I have passed trying to fix this stuff, damn :smile:)

### DIFF
--- a/src/adlmidi_midiplay.cpp
+++ b/src/adlmidi_midiplay.cpp
@@ -1093,8 +1093,7 @@ bool MIDIplay::realTime_NoteOn(uint8_t channel, uint8_t note, uint8_t velocity)
 
     //uint16_t i[2] = { ains.adlno1, ains.adlno2 };
     bool pseudo_4op = ains->flags & adlinsdata::Flag_Pseudo4op;
-    size_t chans_count = (pseudo_4op || (ains->adlno1 != ains->adlno2)) ? 2 : 1;
-    MIDIchannel::NoteInfo::Phys voices[2] =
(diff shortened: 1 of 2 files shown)


Wohlstand:
Thanks for a fix, and yeah, I did right that I made `stable` branch for @loki666 , or this crash would be a painful thing, and I hope, this will no more happen :fox_face: 
I'll check out the stuff at home, for now I'm at my job. Anyway, the code is not so hard and easy to understand. Original Joel's code is a plate of spaghetti I have ate, splited, refactored to make it work better and be more understandable, however some parts are still needs to be improved, for example, moving MIDI Sequencer into separated class to dont mix RT with MIDI Sequencing which now is done. MIDI sequencer made here is powerful and functional, and I gonna use it over FluidLite, Timidity, and over NativeMIDI in SDL Mixer X to provide functionality which is lacking in those MIDI outputs.
However, yeah, you understand this library well, as you making your changes quick and good :smile:

jpcima:
> Original Joel's code is a plate of spaghetti I have ate, splited, refactored to make it work better and be more understandable

My respects for achieving that. I have tried engaging in the same enterprise, but the code has discouraged me.
You make a code which is excellent and clear. The latter is difficult to say of Joel's. :smile: It works well but, it's exactly like you say.

PR #73, opened 2018-04-19
hard-RT: elimination of the user map

jpcima:
Prefer this PR to #69 
It's the same content but rolled in a single commit instead of series of bug fixes..


Wohlstand:
Will review / test this at my home, good job :fox_face:  :+1:

Wohlstand:
Checked, everything works fine. Neat! :fox_face: :+1:

PR #74, opened 2018-04-20
fix implementation defined behavior

jpcima:
It's implementation-defined whether >> is logical or arithmetic on negative integers.
Force a logical shift by casting.


jpcima:
PS The note offsets output by the bank editor are potentially impacted by this..
It would be wise to check.

Wohlstand:
Good thing the bank editor has the unit test with random crap generation which is used to validate by compare odd source and saved then loaded back file data :smile:

jpcima:
Good :+1:

PR #75, opened 2018-04-21
wopl robustness fix

jpcima:
The loader has a byte intended to hold the null terminator of the name.
`snprintf` does not write this byte if the name takes the maximum size (32).

In the bank loading case it's benign because the memory is initialized with zero.
Not in `WOPL_LoadInstFromMem` where the user provides the structure.
Anyway I correct both cases.

---
In my variant of code I will just remove this byte to save the padding space. I am used to handle strings in binary structs, I will not make use of this safety.


Wohlstand:
I think, yeah, even I doing `calloc` to allocate memory with auto-zeroing in the bank loader, it's still be a danger of giving crap-filled buffer from single-instrument loader. Thanks for a catch! :fox_face:

PR #80, opened 2018-04-22
implement copy for OplChannel, and move implementations in .cpp

jpcima:
Same as in libOPNMIDI, including fixes added later

### DIFF
--- a/src/adlmidi_midiplay.cpp
+++ b/src/adlmidi_midiplay.cpp
@@ -2434,17 +2434,7 @@ uint64_t MIDIplay::ChooseDevice(const std::string &name)
 
     size_t n = devices.size() * 16;
     devices.insert(std::make_pair(name, n));
-
-    size_t channelsBefore = Ch.size();
-    size_t channels = n + 16;
-    Ch.resize(channels);
-
-    for(size_t ch = channelsBefore; ch < channels; ++ch) {
-        for(unsigned i = 0; i < 128; ++i) {
-            Ch[ch].activenotes[i].note = i;
-            Ch[ch].activenotes[i].active = false;
-        }
-    }
+    Ch.resize(n + 16);
     return n;
 }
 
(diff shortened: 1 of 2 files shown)


Wohlstand:
I think, I'll wait when you will complete backporting your changes from https://github.com/Wohlstand/libOPNMIDI/pull/21 which is now works perfectly everywhere: opnmidiplay, VLC and phone!

jpcima:
The backporting is all complete. I created this after the final fix of OPNMIDI.
Only tests are missing. You intend to have a copy a ADLMIDI too?

Wohlstand:
Unit tests, I think, would be useful for ADLMIDI too, yeah, I created them on OPNMIDI as I caught a strong crash in it first and I had to debug a library on the phone

Wohlstand:
Gotta to test this around...

jpcima:
It's good to hear the things seem to work at last.
btw: your VLC plugin source is not compatible with my plugin API version. I think it has the code, but commented, and not updated to use `generateFormat`.

I wrote you an issue about huge RAM usage of catch unit tests.
Please consider. [OPNMIDI #19](https://github.com/Wohlstand/libOPNMIDI/issues/19)

Wohlstand:
Just now tested this on my phone, and it works perfectly! :smile:

Wohlstand:
> I wrote you an issue about huge RAM usage of catch unit tests.
Please consider. OPNMIDI#19

Yeah, catch-mini is a choice, even to just don't wait when super-heavey headers tonn will be pased when I really don't need to much crap which is dead code because of unusage. :smiley: 

> btw: your VLC plugin source is not compatible with my plugin API version. I think it has the code, but commented, and not updated to use generateFormat.

Yeah, as I needed to use `libvlc_version.h` and polish the support for newer VLC versions to let it work at you, I mean in VLC 3.0 and higher. My system has VLC 2. However, I built VLC3 from the sources, however, without of GUI, but I only needed for the library headers set and testing backend, so, I have that I needed :fox_face: :smile:

PR #81, opened 2018-04-23
add simple CMakeLists for android

loki666:


Wohlstand:
Why not to have same in-root CMake be used for Android target too? Anyway okay, I'll take this and later I'll make my root CMake script support Android builds.

loki666:
wait I'm testing the root CMakeLists

Wohlstand:
Oh, okay then :wink: 
I'm at my job myself and do nothing with anything here. I usually doing anything at evenings or in days off.

loki666:
yup, root CMakeLists is working fine

Wohlstand:
Anyway, for any Android builds is need to prevent to build console tools and GenAdlData as there are incompatible desktop-only.

loki666:
yes but by default they are disable
one has to know what he's doing...

you can remove the android folder completely if you want, as CMake is the official way to build native suff on android.

Wohlstand:
I mean just mask by
```cmake
if(NOT ANDROID)
option(SOMEOPTION ....)
endif()

if(NOT ANDROID AND SOMEOPTION)
# ...
endif()
```
to disallow user to enable those options on unsupported platform, or alternatively, spawn the fatal message:
```
if(SOMEOPTION)
    if(ANDROID)
        message(FATAL_ERROR "SOMEOPTION is not supported on Android!")
    endif()
endif()
```

loki666:
that's how I'm currently using it

### INCLUDE libADLMIDI LIBRARY ###

```
# Get the absolute path to the adlmidi-lite library directory
get_filename_component(ADLMIDI_DIR
                       ${distribution_DIR}/libADLMIDI
                       ABSOLUTE)

set(libADLMIDI_STATIC OFF CACHE STRING "")
set(libADLMIDI_SHARED ON CACHE STRING "")

add_subdirectory(${ADLMIDI_DIR} ./libADLMIDI)
```

PR #84, opened 2018-05-08
fix pitchbend

jpcima:
Negative pitch bend is broken in one of the function variants.
The one which takes the 14 bit value.


Wohlstand:
Yeah, thanks for a fix!  :fox_face: :+1: 
I wanted to fix that myself, but I was so lazy and worked on different things...

PR #85, opened 2018-05-10
meaningful handling of pitch bend sensitivity

jpcima:
This ADLMIDI code handles the pitchbend sensitivity RPN setting wrong.
It defines `bendsense` according to the RPN MSB, but the `SetRPN` will allow it to take a value from either LSB or MSB, which is an error.

Moreover:

I have found this [definition](http://www.lim.di.unimi.it/IEEE/MIDI/SOT1.HTM#Pitch) of this RPN.
MSB is supposed to be semitones, and LSB in cents.
I add a piece of code which adds consideration for LSB in the computation.

**But**

I've had issues with bizarre sounding pitchbend every now and then, and I try to find why. It's still the case with this patch applied.
For example [this one](http://www.rppmf.com/dream_theater/2005-octavarium/08-dream_theater_2005-octavarium.mid) gets very happy with pitchbend at around 12:15. Timidity is able to get the tone right, but not ADLMIDI.

### DIFF
--- a/src/adlmidi_midiplay.cpp
+++ b/src/adlmidi_midiplay.cpp
@@ -2358,8 +2358,14 @@ void MIDIplay::SetRPN(unsigned MidCh, unsigned value, bool MSB)
     switch(addr + nrpn * 0x10000 + MSB * 0x20000)
     {
     case 0x0000 + 0*0x10000 + 1*0x20000: // Pitch-bender sensitivity
-        Ch[MidCh].bendsense = value / 8192.0;
+    {
+        if (MSB)
+            Ch[MidCh].bendsense_msb = value;
+        else
+            Ch[MidCh].bendsense_lsb = value;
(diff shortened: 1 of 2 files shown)


Wohlstand:
Yeah, it's seems Joel's fault as code related to pitch bend calculation I wasn't changed, and this must be a good fix! :fox_face: :+1: 
Seems OPNMIDI needs to take the same fix as the MIDI Player's code is same.

Wohlstand:
P.S. Nice music you just sent as example! :3

jpcima:
>P.S. Nice music you just sent as example! :3

:wink: 

Yes, but if you have any idea if the tone is wrong, I would love to know.
Making the fix for OPNMIDI

jpcima:
I'm an idiot :man_facepalming: it's not a good fix
There will not be an adverse effect though.
Fixing now for real

Issue #59, opened 2018-04-13
Add the way to change single instrument of currently-loaded bank on the fly.


Wohlstand:
This will allow to use libADLMIDI directly how for OPL3 Bank Editor's backend to change instruments on the fly to test them, or give the ability to use libADLMIDI as VTS-plugin for the musical software.

* - [x] Stop to clean/resize the dynamical instruments data arrays. Instead, override them.
* - [x] Instread of direct access to the generated database, copy instruments of currently choosen embedded bank into dynamical array which is easy to patch.
* - [x] Make instruments store be pre-allocatable which will allow to avoid any reallocations in the real-time when sending instruments on the fly into over-GM range.
* - [x] Add the instrument structure (probably, copy-paste from wopl_file.h from the OPL3 Bank Editor the single-instrument structure) and the function to receive it's bank MSB/LSB, Patch number and type of the instrument (melodic or percussion).

jpcima:
Subscribing myself to thread.
Yes the instrument structure should be exposed.
If it's the wopl directly, even better. I'll use wopl as the internal format to save and restore plugin states.

jpcima:
To add to this, you surely don't need to always copy the bank to dynamic buffer.
The bank pointer count point to either the embedded or dynamic set.
And allocate the dynamic array only as needed. It can be copy-on-write.

>and the ID of the channel.

Why does the channel relate to instrument definitions? is it not the library which does the channel mappings internally according to program change events?

Wohlstand:
> and the ID of the channel.

Oops, it's totally nosense... I fixed it with `type of the instrument (melodic or percussion).`

jpcima:
I started the foundation for a music plugin. Development will be [here](https://github.com/jpcima/ADLplug).

jpcima:
After hard-rt is finished which is very soon, this bug will be next target. :dart: :gun:
You have a status update on this one? no matter if there isn't, I just don't want to get in your feet if I'm starting something.

jpcima:
It was nice of you to edit the title, but I think you forgot to answer my question. :smile: 
I was asking if you have made some progress on this issue already.

Wohlstand:
[...]

PR #87, opened 2018-05-15
specialized hash table for bank number mappings

jpcima:
#59 

I create a hashmap container which is specialized for mapping things to bank numbers.
This replaces the current `BankMap`. A unit test is included.
This structure is optimized for worst case, because it's not only easier but also for RT reasons, and it allows for some cache locality (if it was reserved in one block).

The standard containers would not provide the functionality needed, because they do not allow to reserve storage for a fixed number of elements. I have a normal insertion which follows STL semantics, and a policy-based insertion which does not attempt to extend the storage.

Now, I am thinking about storing the instruments in this structure instead of having the indirection into instrument arrays of ADLMIDI. It's because the erasure operations are going to create holes, and the operations may require worst case in O(N). (and anyway, it will not be any worse than midiplay currently, which does its std::map lookup every NoteOn).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -488,6 +488,12 @@ install(FILES
         include/adlmidi.hpp
         DESTINATION include/)
 
+option(WITH_UNIT_TESTS "Enable unit testing" OFF)
+if(WITH_UNIT_TESTS)
+    enable_testing()
+    add_subdirectory(test)
+endif()
+
 message("==== libADLMIDI options ====")
 message("WITH_MIDIPLAY            = ${WITH_MIDIPLAY}")
 message("WITH_VLC_PLUGIN          = ${WITH_VLC_PLUGIN}")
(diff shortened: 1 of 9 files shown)


Wohlstand:
Also, replacing of standard containers here would help me to check out the thing on Wattcom thing again, as previously I have built the thing, but it failed to work due strong bugs in a stack processing I had found while long research and analytics: https://github.com/open-watcom/open-watcom-v2/issues/391

Wohlstand:
Damn! You forgot one thing that code must be in C++98 because of Android SDK and older compilers support

Wohlstand:
So, now I trying to fix something by myself... And, std::unique_ptr I'll replace with existing custom-made pointer class I'll move into separate header for common usage in different headers

Wohlstand:
And another thing, using of unique pointers is no way use the std::list container as it tries to use Copy constructor. I only can make it work as move constructur

jpcima:
> as it tries to use Copy constructor

I don't think so. Values in a std::list do not have to be moved, unlike in std::vector in case of growth.

> Addition, removal and moving the elements within the list or across several lists does not invalidate the iterators or references. An iterator is invalidated only when the corresponding element is deleted.

Wohlstand:
Okay, as now I'm tired, I have posted my small changes into "WIP-banks" branch you can pull to you.
What you need to do:
* Avoid in-class assignments, move them into constructor. C++11 allows that, but not C++98.
* Don't use any std::* smart pointers and use/polish custom-made shared pointer things.
* Don't use `nullptr`, use `NULL` with no matter to warnings. `nullptr` is C++11 only.
* I think, you better use something different from the std::list as unique pointer concept is not so good for that...

jpcima:
I will put the compiler in std=c++98 and I'll fix everything. See ya.

PR #88, opened 2018-05-16
c++98 support for bank map

jpcima:
C++98 bank map implemented using custom shared pointers.
The shared pointer receives a deleter argument, which is to define in order to invoke `delete[]` on the pointer which is wrapped.
Also changed the more minor things here and there.

### DIFF
--- a/src/adlmidi_bankmap.h
+++ b/src/adlmidi_bankmap.h
@@ -25,9 +25,8 @@
 #define ADLMIDI_BANKMAP_H
 
 #include <list>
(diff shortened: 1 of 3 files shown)


jpcima:
I have cleaned up the smart pointers which have become too big of a mess.
There is more generic usage, the redundant stuff is removed.

Wohlstand:
Okay, as this branch now matching one of older states of your `banks2` and now it is also merged with master, I'll nuke this branch at me

PR #89, opened 2018-05-16
Unified structure for instruments and meta-instruments

jpcima:
Rebased on `WIP-banks` branch which adds c++98. The important commit to look at is https://github.com/Wohlstand/libADLMIDI/pull/89/commits/90d76408d83e07e43dbea404ddbb94c0f6305b24.

This adds `adlinsdata2` which is a structure having operator data inside, instead of the numeric references `adlno1` `adlno2`. This can be the new instrument structure to keep them stored in one piece.
Internally, I edit libADLMIDI to use this structure.

But I do not edit `genadldata` and allow it to convert to the same structure as before, because it is more compact in memory. I keep both structures available. There exists a constructor to convert `adlinsdata` to `adlinsdata2`.

There is one impact on implementing libADLMIDI : wherever the `adlno`s are compared for equality, I am now instead comparing the `adldata` structures entirely (the 11 bytes).

ADDENDUM
* for adlmidi_load, I remove the intermediate structure `WOPL_Inst` which is not needed, because `adlinsdata2` is usable for the same thing

### DIFF
--- a/src/adlmidi_load.cpp
+++ b/src/adlmidi_load.cpp
@@ -72,59 +72,50 @@ bool MIDIplay::LoadBank(const void *data, size_t size)
     return LoadBank(file);
 }
 
-
-struct WOPL_Inst
+static void cvt_WOPLI_to_FMIns(adlinsdata2 &ins, WOPLInstrument &in)
 {
-    bool fourOps;
-    char padding[7];
-    struct adlinsdata adlins;
-    struct adldata    op[2];
-    uint16_t ms_sound_kon;
-    uint16_t ms_sound_koff;
-};
-
-static void cvt_WOPLI_to_FMIns(WOPL_Inst &ins, WOPLInstrument &in)
-{
-    ins.op[0].finetune = in.note_offset1;
-    ins.op[1].finetune = in.note_offset2;
-
-    ins.adlins.voice2_fine_tune = 0.0;
+    ins.voice2_fine_tune = 0.0;
     int8_t voice2_fine_tune = in.second_voice_detune;
     if(voice2_fine_tune != 0)
     {
         if(voice2_fine_tune == 1)
-            ins.adlins.voice2_fine_tune = 0.000025;
+            ins.voice2_fine_tune = 0.000025;
         else if(voice2_fine_tune == -1)
-            ins.adlins.voice2_fine_tune = -0.000025;
+            ins.voice2_fine_tune = -0.000025;
         else
-            ins.adlins.voice2_fine_tune = ((voice2_fine_tune * 15.625) / 1000.0);
+            ins.voice2_fine_tune = ((voice2_fine_tune * 15.625) / 1000.0);
     }
 
-    ins.adlins.tone = in.percussion_key_number;
+    ins.tone = in.percussion_key_number;
+    ins.flags = (in.inst_flags & WOPL_Ins_4op) && (in.inst_flags & WOPL_Ins_Pseudo4op) ? adlinsdata::Flag_Pseudo4op : 0;
+    ins.flags|= (in.inst_flags & WOPL_Ins_IsBlank) ? adlinsdata::Flag_NoSound : 0;
 
-    ins.adlins.flags = (in.inst_flags & WOPL_Ins_4op) && (in.inst_flags & WOPL_Ins_Pseudo4op) ? adlinsdata::Flag_Pseudo4op : 0;
-    ins.adlins.flags|= (in.inst_flags & WOPL_Ins_IsBlank) ? adlinsdata::Flag_NoSound : 0;
-    ins.fourOps      = (in.inst_flags & WOPL_Ins_4op) || (in.inst_flags & WOPL_Ins_Pseudo4op);
-
-    ins.op[0].feedconn = in.fb_conn1_C0;
-    ins.op[1].feedconn = in.fb_conn2_C0;
-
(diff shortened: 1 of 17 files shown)


src/adlmidi_load.cpp
```diff
@@ -72,59 +72,50 @@ bool MIDIplay::LoadBank(const void *data, size_t size)
...
-struct WOPL_Inst
-{
-    bool fourOps;
-    char padding[7];
-    struct adlinsdata adlins;
-    struct adldata    op[2];
-    uint16_t ms_sound_kon;
-    uint16_t ms_sound_koff;
-};
-
```
Wohlstand: Yeah, it was just a convenient proxy, and it's good when you have made direct passing of the data into final store :+1:

Wohlstand:
Okay, I'll completely check out the thing when I'll reach my home...

jpcima:
This part is good to use I think, after I got this stupid bug fixed which forbid loading from files.
I have a next step which finally makes it store in map and take banks from there. All the code is implemented but I must yet make it work.

Wohlstand:
One only thing I need to rework - while reading EA-MUS or CMF files, convert the data into MIDI events and avoid extra boolean flags are now changing synthesizer behavior for EAMUS or for CMF

Wohlstand:
That now is possible as I already implemented pre-processing of a whole thing to provide validation of loop poi ts and merge different events of same sense into one

jpcima:
From my understanding the instruments are using adlno1==adlno2 to indicate 2op.
Are 4op patches not permitted to have identical operator pairs?
Why doesn't ADLMIDI make this a flag, like in WOPL?

Wohlstand:
if adlno1 and adlno2 are equal for 4-op, then I having the "WOPL_4op" flag to force 4op usige. Bisquid had a bug with this I have been fixed. Same crap was on pseudo-4op I also fixed when two voices are equal.

jpcima:
This flag `WOPL_Ins_4op` does not get translated into ADLMIDI instrument structures (unless pseudo).
From only `adlinsdata`, I mean it's impossible to distinguish a 2op from a natural 4op.
All you know is that `adlno` are equal or not, and it's used as a hint for considering 2op.
And there is a number of `adlno1==adlno2` in code (or my case, `adl[0]==adl[1]`), which are basing their guess on this.

jpcima:
You have any examples of Xg percussions files to test with?

jpcima:
Thank you. I will try. I have been confused by the percussion stuff a bit but I'm on a good track.

Wohlstand:
On a quick hand I have applied few commits and posted as `banks2` branch, pull them to you :fox_face: :wink:

Wohlstand:
XG way to turn channel into percussion is much easier by back number tricks, the GS way requires SysEx evetns support I wasn't implemented yet

jpcima:
Ok I have pulled these commits. I don't have problems with `banks3` currectly and the percussion seems right.
If this is correct, there should remain only to stick an API on top of it and call it done.

Wohlstand:
Anyway, as I have very outdated OpenWatcom 2.0 toolchain I have the crap with `CopySamplesTransformed` template:
default
I need to update OpenWatcom toolchain and retry this thing, and if it's still be, I'll report them it if someone wasn't reported this crap before...

jpcima:
Does it help to specify template argument 2 of CopySamplesTransformed? instead of infering

Wohlstand:
And yeah... it happens on a double/flat, but not on integers...
default
What if I will change some....

jpcima:
The difference I see is `adl_cvtReal` being an instantiation of template.
I would try defining `adl_cvtFloat` and `adl_cvtDouble` and see if that changes a thing.

Wohlstand:
I made this, and crap has gone:
default

But, another crap became: 
default

which was been "fixed" by this:
default

jpcima:
You are the only programmer I ever knew who has a developed interest for vintage compiler toolchains :smile:

Wohlstand:
Vintage compilers are usually for fun and history curiosity :wink: 
I once ported one my simple WinAPI program written to be run on WinXP, into Win 3.1 and built with Visual C++ 1.2 that ran in 16-bit mode, and I have found what's the "previous" context was in WinAPI apps. In modern Windows it's nonsense, but on Windows 3.x it's very important as a way to connect that already-running window back and don't run duplicated apps.

Wohlstand:
At the same time, it still won't work through OpenWatcom, but it works fine under DJGPP, so, no troubles :wink: :fox_face: :+1:

Wohlstand:
Anyway, I'll drop those small changes I made at me as that was a small experiment and I'll try to tweak that a bit later, for now I'm tired and gonna sleep now...

jpcima:
One of the things I enjoy in old schools computing is Common Lisp, not many things outside of that.
c++ below 11 for me, is no thanks. I've used c++0x in a r&d job since ever available, in the experimental stages. :smile:

Wohlstand:
As I came to my home too late, therefore I can't just test the stuff now... So, gotta to do that tomorrow when I'll have two free days (Saturday and Sunday). I wanted to try the stuff out on my phone to catch possible crap around if it was missed up, but when I tried some basics yesterday, the stuff is seems fine :+1: 
However, I still need to test out the stuff more deep...

jpcima:
Take all your time. For now I work in my own branch and I continued work in the VST by implementing user code there. I should add ability to properly load banks in no time now, and then add parameter controls in the UI.

PR #90, opened 2018-05-16
make the AdlInstrumentTester a P-Impl

jpcima:
I will modify `AdlInstrumentTester` for the incoming bank work.
In C++ unfortunately, changing the fields of a class is an ABI breakage.
I am redesigning this class for P-Impl use to give it stable ABI; it's the same pattern that Qt libraries use.

### DIFF
--- a/utils/adlmidi-2/midiplay.cc
+++ b/utils/adlmidi-2/midiplay.cc
@@ -1758,7 +1758,7 @@ int main(int argc, char **argv)
 
     if(argc >= 3)
     {
-        if(is_number(argv[2]))
+        if(is_number(argv[2]) || !strcmp(argv[2], "-1"))
         {
             int bankno = std::atoi(argv[2]);
             if(bankno == -1)


jpcima:
I don't understand how to use this command line. Have you used it yourself?

It looks for "-1" passed as bank number. But before, the function `is_number` is checking for `isdigit` on all the characters, which '-' is not. There has to be a problem here.

Wohlstand:
You only need to feed any MIDI file (yes, it's a clunky API) and yes, put the -1 as a bank number, or even pass one of special flags... I forgot myself too... But I successfully ran the instrument testing once upon...

Wohlstand:
The "atoi" must detect "-1" as it is designed for signed integers

jpcima:
It responds to keys like it seems it should.
(this terminal UI however still displays like crap to me. It was already the case with Bisqwit ADLMIDI)

Wohlstand:
The displayed crap was an old thing,  and yeah, why not to completely rework it? I mean make the whole UI of it to be more robust and beautiful? Anyway it's not so useful as this ADLMIDI2 is intended to match most of Bisquit's original in most. At me is also crap, and because of UI are strong lags on Windows due it's super slow console printing responsing

jpcima:
I would like if there were a library which permits some TUI programming like curses, but at the same time which I could embed in JUCE graphics, and mix with normal drawn graphics.

It's not out of question that I derive a program from my current curses or JUCE interface, and I transform it into a MIDI player. I'll think more about this after completing the synthesizer work.

I like the Adlib Tracker, it's a nice use of terminal graphics for user interface.

Wohlstand:
Anyway, I'll take this now

PR #94, opened 2018-05-31
a small mistake fixed about maximum note number

jpcima:


Wohlstand:
Oops... But thanks :fox_face: :smile:

Issue #92, opened 2018-05-25
Some notes can be cut even if there are many, many channels available


sneakernets:
This one affects "They're Going To Get You" from Doom. The high-pitched drone string "cuts out" even if there are ample channels available. In addition to that, the pizzicato strings "pop" as soon as the drone string is silenced. Very weird.

I have supplied a 25 second OGG of this happening. Sorry for this being loud, but I amp'd it up so you can hear. Happens between 20-21 seconds into the recording. increasing the number of chips to an obscene number "fixes" it, but it makes my recording software lag something fierce, so that's probably not a good remedy for it.
[getyou.zip](https://github.com/Wohlstand/libADLMIDI/files/2040743/getyou.zip)

Wohlstand:
I think, I'll need to make the monitor of busyness of channels to look how many channels are busy in a real. And yeah, for a long notes (are don't stopping while key is held), I must make the protection from a kill and kill channels are already released...

papiezak:
I also spotted note cutting bug, however much more extreme than that:
[note cut.zip](https://github.com/Wohlstand/libADLMIDI/files/2042570/note.cut.zip) - flute section is almost mute, compared to recording of what should it be. Midi file included. Every bank affected, no matter whether 2-op or 4-op, both OPL3 and OPN2.

Wohlstand:
With this: https://github.com/Wohlstand/OPL3BankEditor/pull/71 and with this https://github.com/Wohlstand/libADLMIDI/commit/70c83caace9911dc5d8b80c1653c96b073f88e4a the issue is no more reproducing on your bank even with 4 chips! :fox_face: :+1: Tremolo-strings still persistent and no more dying!
I still to do some works until I'll apply it to everywhere...

Wohlstand:
[getyou-fixed.zip](https://github.com/Wohlstand/libADLMIDI/files/2107648/getyou-fixed.zip)

papiezak:
yeah, looks like fixed now.
midi that I linked few comments above still sounds broken, but it's most likely not connected to this issue; frankly speaking I have no bloody idea why it's that.

Wohlstand:
As it's measurer problem in most, you need to download super-fresh version and run the full re-calculation to have that fixed. Embedded banks are NOT fixed yet as I wasn't ported fresh measurer into GenAdldata yet.

papiezak:
[...]

PR #95, opened 2018-05-31
attempt to rework 4op count computation

jpcima:
#92 

I would like to have a discussion on computation of 4op channel count.
It's a code I don't like, for the hazardous way to statically fix the channel count, and also I find it to be nonsensical in several ways.
I know that this thing was Joel's code originally.

1. The dynamic banks counts the blank instruments, which makes total count erroneous for the final estimation. It's a thing I have attempted to fix.

2. There is a modulo 128 applied to the total I find no explanation for, and which I would think is better to remove.

3. I understand the computation of `NumFourOps` like this:
- If there's no 4-ops => 0
- The 4-ops use more than 7/8 of the bank => 6
- The 4-ops use less than 1/8 of the bank => **0** (**can't have any 4-ops in this case??????**)
- Otherwise if it's single chip 1, else 4

The error message following makes no sense. If there is >= 15/16 instruments with 4-op, `NumFourOps` will not be 0, so how can the condition ever be entered?
Beside, ADLMIDI never checks the return value of this function, so this error condition is completely meaningless in context.

### DIFF
--- a/src/adlmidi_private.cpp
+++ b/src/adlmidi_private.cpp
@@ -36,17 +36,15 @@ int adlRefreshNumCards(ADL_MIDIPlayer *device)
         //For custom bank
         for(size_t a = 0; a < play->opl.dynamic_metainstruments.size(); ++a)
         {
+            adlinsdata2 &ins = play->opl.dynamic_metainstruments[a];
+            if(ins.flags & adlinsdata::Flag_NoSound)
+                continue;
+
             size_t div = (a >= play->opl.dynamic_percussion_offset) ? 1 : 0;
             ++n_total[div];
-            adlinsdata2 &ins = play->opl.dynamic_metainstruments[a];
             if(ins.flags & adlinsdata::Flag_Real4op)
                 ++n_fourop[div];
         }
-
-        play->m_setup.NumFourOps = ((n_fourop[0] == 0) && (n_fourop[1] == 0)) ? 0
-                : (n_fourop[0] >= (128 * 7) / 8) ? play->m_setup.NumCards * 6
-                : (n_fourop[0] < (128 * 1) / 8) ? (n_fourop[1] > 0 ? 4 : 0)
-                : (play->m_setup.NumCards == 1 ? 1 : play->m_setup.NumCards * 4);
     }
     else
     {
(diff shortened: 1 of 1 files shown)


jpcima:
The driver of J.A.Nguyen had dynamic allocation of channels. Can't ADLMIDI have ability to do this too?

Wohlstand:
Do you mean toggling on/off of 2op channels on a fly on taking usage of 4op instruments? It's looks a bit danger in a case when used lot of 2op channels and needed to use some 4op channels are trying to steal two 2op channels (per each 1 4op) are busy on a moment...

jpcima:
Dynamic allocations means that notes of 4op instruments will be allowed to take all the channels as they want as long as there is room.

When you look at it for musical usage like VST, an allocation like this makes more sense.
A musician is maybe going to just use the player for a little number of instruments.

Playback of MIDI files is different where you want to optimize to fit all instruments on one set of chips. And it's what this numbre of card computation tries to optimize for.

But me, when I will program the VST, I don't care. But it's not necessary that channel count must be full dynamic. In fact, what I plan to do, is that I will have a control in UI to let user decide the channel count.
https://github.com/jpcima/ADLplug/issues/3

Wohlstand:
> Dynamic allocations means that notes of 4op instruments will be allowed to take all the channels as they want as long as there is room.

This is directly means when in a time of playing of any 2op instruments, first 4op instrument introduced will nuke some of 2op instruments to release channels to take one 4op channel.

As my MIDI sequencer has the preprocessor, then it can check the complete MIDI file for any path change events and then analyze using instruments for 4op and allocate necessary 4op channels count especially for this loaded MIDI file. In case of real-time, then is no way to guess how many 4op instruments musician will use now and that yeah.

For the dynamical switch of 4op channels count it's need to improve the function and do killing of notes are using channels are needed to be switched. Means, when you toggling on some 4op channels, you need to detect two notes are using 2op channels are needed to allocate this 4op channel, and only in the case when is no free 2op channels are can be used for allocation of 4op channel without of note kills. For now the 4op channels counter kills everything on every call.

jpcima:
> For the dynamical switch of 4op channels count it's need to improve the function and do killing of notes are using channels are needed to be switched.

It's not important, it's not a setting which is going to change every second.

Wohlstand:
> It's not important, it's not a setting which is going to change every second.

If you mean manually only, then it's another thing. I though you gonna to make the fully-automatic thing that will toggle count of 4-op channels on necessity...

jpcima:
> I though you gonna to make the fully-automatic thing that will toggle count of 4-op channels on necessity

No there are better things to focus on right now

Issue #97, opened 2018-05-31
Potential issue in Dosbox OPL code


jpcima:
There is a very high probability of presence of a bug in Doxbox OPL.
This is to verify. This problem is reported by a recent GCC compiler:

Take a look at this `else` branch:

https://github.com/Wohlstand/libADLMIDI/blob/649c9dcc2ba53a0be994318999beb905c0c14528/src/chips/dosbox/dbopl.cpp#L1427-L1431

And the macro definition of `REGCHAN`

https://github.com/Wohlstand/libADLMIDI/blob/649c9dcc2ba53a0be994318999beb905c0c14528/src/chips/dosbox/dbopl.cpp#L1360-L1365

As you can see, a part of this code is not covered under `else` when the macro is expanded.
It is to be checked what impact that bug has possibility to create.

jpcima:
Fixed in upstream Dosbox btw. An update is probably a good idea.

https://sourceforge.net/p/dosbox/code-0/HEAD/tree/dosbox/trunk/src/hardware/dbopl.cpp

Wohlstand:
This DosBox emulator here is VERY outdated and it needed to be updated, yeah!  :3
So, it's only need to take the latest version and polish it a bit if it has some missing functions are customy made...

jpcima:
Look what just happened. Tadaaa! :smile:

Wohlstand:
as DosBox emulator has been updated and this bug is no more a problem, I closing this :fox_face:

PR #98, opened 2018-05-31
update Dosbox OPL

jpcima:
#97 

Update Dosbox to SVN trunk

I change all things necessary to make it build and include the modifications from Joel.
- decouple from Dosbox core
- include in it the needed routines of ChipBase, and the ability to generate adding


Wohlstand:
So, It's needed to repeat a fix of MSVC build which now is broken (yeah, DBOPL originally was incompatible to MSVC, but I had to fixed the build, and yeah, it's need on my side to repair the support). So, gotta to debug...

PR #99, opened 2018-06-01
keep the 4op count low for single chip

jpcima:
I think you have forgotten to take this into account for the formula.
You have mentioned how this was important especially in a hardware situation.

### DIFF
--- a/src/adlmidi_private.cpp
+++ b/src/adlmidi_private.cpp
@@ -76,13 +76,19 @@ int adlRefreshNumCards(ADL_MIDIPlayer *device)
     else if(n_fourop[0] > 0)
         numFourOps = 4;
 
+    // Scale to chip count, but keep the 4op usage low on single chip
+    if(play->m_setup.NumCards > 1)
+        numFourOps *= play->m_setup.NumCards;
+    else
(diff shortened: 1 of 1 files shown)


src/adlmidi_private.cpp
```diff
@@ -76,13 +76,19 @@ int adlRefreshNumCards(ADL_MIDIPlayer *device)
     else if(n_fourop[0] > 0)
         numFourOps = 4;
 
+    // Scale to chip count, but keep the 4op usage low on single chip
+    if(play->m_setup.NumCards > 1)
+        numFourOps *= play->m_setup.NumCards;
+    else
+        numFourOps = (numFourOps > 0) ? 1 : 0;
```
Wohlstand: This will mean EVERYONE will use single 4op on single-chip...

Wohlstand:
If we have 18 2ops and no 4ops, but having 2 4ops for drums, we will still have 14 2ops that must be fine. I have tested my melody in DJGPP-built player and channels management now is correct

jpcima:
Ok. I wanted to integrate the new Dosbox in adljack and I had a suspicion a thing was forgotten here.
If it's fine, I don't have any more things to say about it.

Issue #43, opened 2018-03-26
Implement "blank" instruments support for WOPL


Wohlstand:
As melodic multi-bank behavior was never tested, I have missed up the case when lots of instruments inside of same non-zero bank are blank. There are must fall into the root bank to avoid a silence.
In the OPL3 Bank Editor, add the "No-Sound" flag into WOPL format into each instrument entry.
On side of libADLMIDI, detect this flag on instrument entry when multi-bank is in use and fall back to zero bank.

Wohlstand:
Made!

PR #100, opened 2018-06-01
bank storage inside dynamic map

jpcima:
This is the storage of structure `adlinsdata2` inside the dynamic map.

Now, all bank storage used by the MIDI player happens is made to happen inside this map.
For embedded banks, when `setEmbeddedBank` is called, it will copy the internal bank into entry 0:0 of the map.

The bank identifier is the 16bit integer PMMMMMMM 0LLLLLLL. (Percussive, MSB, LSB)
A bit operation identifies the bank as melodic or percussion. It's the bit value `OPL3::PercussionTag`.

When the MIDI player won't find a bank, il will fall back to 0:0.

I find it not bad as it is now, but I think later I'll be interested to keep a counter of the number of actual (non-blank) instruments contained in the banks slots.

### DIFF
--- a/src/adlmidi_bankmap.h
+++ b/src/adlmidi_bankmap.h
@@ -27,6 +27,7 @@
 #include <list>
 #include <utility>
 #include <stdint.h>
+#include <stddef.h>
 
 #include "adlmidi_ptr.hpp"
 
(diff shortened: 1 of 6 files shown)


Wohlstand:
About of the fallback: there are not banks only, but some blank instruments in one bank are also must fall back to 0:0, look the issue https://github.com/Wohlstand/libADLMIDI/issues/43 , once in some place Sneakernets have reported that some musics are using non-zero bank and catching blank voices, doesn't fall backs to zero bank.

I'll review the whole stuff at evening :wink:

jpcima:
It's a good remark, and one quickly fixed.
It's a separate commit for readability because it grows the diff a bit for not much reason.

PR #101, opened 2018-06-02
dynamic instrument API

jpcima:
#59 

This is a simple API for external access of dynamic banks.

Most operations are here, except some I will be interested to add in a moment:
- creating a bank with possibility to reuse a blank bank slot of there exists one. (to add as `ADL_BankAccessFlags`) [1]
- getting the number of banks. I don't have a lot of use for this, and this would enter in naming confusion with the unfortunately existing `adl_getBanksCount`.

To accelerate [1], I wish to keep a counter of used instruments in dynamic banks.
It would be initialized when a bank is loaded either from embedded or WOPL file source, and successively incremented or decremented by following calls of `adl_setInstrument`.

### DIFF
--- a/include/adlmidi.h
+++ b/include/adlmidi.h
@@ -111,6 +111,46 @@ extern int adl_getBanksCount();
 /* Returns pointer to array of names of every bank */
 extern const char *const *adl_getBankNames();
 
+/* Reference to dynamic bank */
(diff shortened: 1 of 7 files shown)


Wohlstand:
Anyway, when I will rework rhythm-mode drums, then is need to give the support for them into enums

PR #103, opened 2018-06-03
fix build under dynamic bank map and embedded banks disabled

jpcima:


Wohlstand:
Thanks! :fox_face: :+1:

Wohlstand:
But yeah, I probably must to add more matrix branches to include disabled/enabled some of macros like this

papiezak:
Yeah, it's good now. Thanks!

P.S: nukedopl3.c, line 1150:
`if (chip->eg_timer == 0xfffffffff)`
compiler cries that integer constant is too large from long type

jpcima:
Yeah true, this constant needs a "U" for unsigned, ie. `0xfffffffffU`
It's harmless though

Wohlstand:
The one thing is C90 mode is turned on, and it's unsigned Long Long on 32-bits while regular "long" is 32-bits on 32-bit and on 16-bit systems.

PR #104, opened 2018-06-03
optional high-quality resampling

jpcima:
Add a HQ resampling option to be used with Nuked OPL.
This will allow to hear a comparison for high sample rates, to compare with original linear interpolation.

Performance-wise, the overhead of resampling is a 48-tap FIR on each channel.
There is a description of the resampling process: http://kokkinizita.linuxaudio.org/linuxaudio/zita-resampler/resampler.html


Wohlstand:
Neat!
It's need to apply it to OPNMIDI to let MAME use YM2612's native sample rate to give sounding that will match real chip or Nuked OPN2 :wink: 
Gotta to check out this in work...

jpcima:
Good. Don't expect a lot of miracles to come out of this change of resampling.
I gave a it a few tries before and I didn't find too significant differences. But you can look in Audacity, you'll see it reshaped the signal quite a bit in observable ways.

I expect the most important effect to be an increase of CPU use. For this, I promised myself lots of time ago to rework the convolution in the zita library to SIMDize it for a 4x or 8x speed gain, and it's a thing that I still haven't done.

PR #106, opened 2018-06-04
new pitchbend strategy

jpcima:
Proposing alternative to the pitchbend fix I'm in disagreement with.

Instead of updating pitch when bend sensitivity events are coming, I do this:
Do NOT premultiply pitchbend with bendsense when the pitchbend event arrives.
Conserve the midi value stored, and wait in order to do this multiplication late, at the time of note-on.
What do you think?

~~Still, I'm not convinced if resetting bend on `resetAllControllers` is a proper thing to do.~~
EDIT it's the good thing to reset pitch bend on this event

### DIFF
--- a/src/adlmidi_private.hpp
+++ b/src/adlmidi_private.hpp
@@ -504,8 +504,8 @@ public:
         //! Is note aftertouch has any non-zero value
         bool    noteAfterTouchInUse;
         char ____padding[6];
-        double bendSrc;
-        double  bend, bendsense;
+        int bend;
+        double bendsense;
         int bendsense_lsb, bendsense_msb;
         double  vibpos, vibspeed, vibdepth;
         int64_t vibdelay;
@@ -699,8 +699,7 @@ public:
         }
         void resetAllControllers()
         {
-            bendSrc = 0.0;
-            bend = 0.0;
+            bend = 0;
             bendsense_msb = 2;
             bendsense_lsb = 0;
             updateBendSensitivity();
@@ -724,9 +723,8 @@ public:
         }
         void updateBendSensitivity()
         {
-            int cent = bendsense_msb + static_cast<int>(static_cast<double>(bendsense_lsb) * (1.0 / 128.0));
-            bendsense = static_cast<double>(cent) / 8192.0;
-            bend = bendSrc * bendsense;
+            int cent = bendsense_msb * 128 + bendsense_lsb;
+            bendsense = cent * (1.0 / (128 * 8192));
         }
         MIDIchannel()
         {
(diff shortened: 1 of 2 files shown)


src/adlmidi_private.hpp
```diff
@@ -724,9 +723,8 @@ class MIDIplay
         }
         void updateBendSensitivity()
         {
-            int cent = bendsense_msb + static_cast<int>(static_cast<double>(bendsense_lsb) * (1.0 / 128.0));
-            bendsense = static_cast<double>(cent) / 8192.0;
-            bend = bendSrc * bendsense;
+            int cent = bendsense_msb * 128 + bendsense_lsb;
+            bendsense = cent * (1.0 / (128 * 8192));
```
Wohlstand: Okay, THIS formula produces the detuned sound. Just now I have made demo of your formula and the old formula:
[PitchBend-8192-results.7z.zip](https://github.com/Wohlstand/libADLMIDI/files/2067529/PitchBend-8192-results.7z.zip)

jpcima: I'll test this against Fluisynth. What your previous formula has done is simply to cancel the effect of sensitivity LSB.

I uploaded a tool because I lack controller ability in MIDI stuff and it is an impractical situation for test. https://github.com/jpcima/midi-shell
And I will add some XG ability, to be able to control my synthesizer.

jpcima: I have an element of answer for you: Fluidsynth **ignores** sensitivity LSB

jpcima: It's difficult to find any open source synths which implement this LSB. In Code search, I have not been able to find just one. Maybe we should just drop implementing this LSB value.... but I think it's really supposed to be 1/100ths values.

I'll make you an entire archives of MIDIs I have using pitch LSB. (I have lots)

jpcima: Have MIDI files. This thing is long to run on collections and here's part of findings.
It's files which use non-0 pitch sense LSB. You have pitch sense events logged in text files at the root.

[midis1.tar.gz](https://github.com/Wohlstand/libADLMIDI/files/2067924/midis1.tar.gz)

jpcima: All MIDI affected I have in this computer [fullmidi.tar.gz](https://github.com/Wohlstand/libADLMIDI/files/2067963/fullmidi.tar.gz)

Wohlstand: Okay, just now I have tested my PlusDemo on SB-Live's synth, and the sound is same distorted as your demo. So, i'll re-tune my demos and then I'll test them on updated OPxMIDI's code...

PR #108, opened 2018-06-07
library version string for HQ variant

jpcima:
The same modification of the version string as in OPNMIDI.

On a side note: I observe now `genadldata` has a failing link step in the HQ build. Undefined references to zita library symbols.


jpcima:
For a potential solution about the problem of `genadldata` not linking:
After porting to ADLMIDI the chips work of OPNMIDI, there are native rate generation APIs.
So, measurer can just be made to generate at native rate and eliminate the resampling step.

Wohlstand:
Yes, it's a very good idea to just use native rate in the GenAdldata to analyze the sounding... :wink:

PR #109, opened 2018-06-07
chips: create a common method set for generation and resampling

jpcima:
Add the generic chips API with resampler.
Allows Dosbox to generate at native rate (but for now, buffered)


Wohlstand:
It's a good thing to try how it (DosBox) will sound... 

Anyway, on my phone on OPNMIDI I have found a con of a force native rate usage: obvious increasing of CPU usage and my demo began to play choppy sound on MAME with multichip...

Wohlstand:
P.S. Okay, for the note: give the xxxMIDIPlay tool to use different sample rates, and totally rework the arguments system, for example, I'll try to apply the powerful tclap thing I began to use on PGE Project tools and I liked it. :smile:

jpcima:
> obvious increasing of CPU usage and my demo began to play choppy sound on MAME with multichip

The ratio of sample rates: `53267. / 44100. = 1.20` ⇒ 20% increase in computation
So yeah definitely a considerable increase for more limited systems.

It is always worth a study to profile these emulators and see how to enhance the performance.

I have not checked too much, but I have impression in OPNMIDI that differences between emulators are way reduced after the sample rates have been normalized.

papiezak:
>I have not checked too much, but I have impression in OPNMIDI that differences between emulators are way reduced after the sample rates have been normalized.

Yeah, but still Gens core doesn't support SSG-EG properly, ignores distortion caused by YM2612 DAC (I mean clipping at high volumes and obviously no ladder effect emulation), and I have impression that detune sounds differently than in other emus. It's discontinued, after all.

MAME can be be futher updated, though. Since 2017, the team behind it added lots of improvements. Sadly, you can't use newest core since multichip support was removed :face_with_head_bandage: . some features MIGHT be possible to backport, but one would have to check the performance impact. Then, results wouldn't be far off from Nuked OPN2.

jpcima:
@papiezak OPNMIDI has only some minor diffs with MAME master. I assume you must refer to this version out of Genesis-Plus-GX: https://github.com/ekeeke/Genesis-Plus-GX/commits/master/core/sound/ym2612.c

I have found the old conversation which mentions the [multi-chip removal](https://github.com/Wohlstand/OPN2BankEditor/issues/7).

> Without multi-chip will be impossible to use in libOPNMIDI as multi-chip is required and widely used to excite 6 channels limit.

I don't know what is meant by this. Unless I'm wrong, it has, like ADLMIDI, its own management of multiple chip and does not need it provided externally. Maybe it wasn't the case at the time of writing this comment?

Wohlstand:
> I don't know what is meant by this. Unless I'm wrong, it has, like ADLMIDI, its own management of multiple chip and does not need it provided externally. Maybe it wasn't the case at the time of writing this comment?

6 channels is too small count of channels to play complex music on the single chip. OPL3 chip even has 18 2op channels which is fine for most of existing musics. However, it's 4-op mode allows only 6 channels like OPN2 does by default. Multi-chip support in most is needed to increase the polyphony width to be able correctly play complex musics are relying on the wide polyphony usage. In some cases it's bad practice to break playing channels or do "arpeggioing" of chords, which makes playing of music be incorrect. ADLMIDI has introduced a good thing with multichip usage, and I took this for OPNMIDI as I already told, 6 channels limit is too small for a complex music.

jpcima:
I understand this fully already. It's not the point of my statement above.
In old comment threads you have said about emulator-side removal side multichip, that it was a barrier for using in OPNMIDI.
But, why does libOPMIDI usage care about this matter? when like you say, OPNMIDI took the same chip management as ADLMIDI already had.

Wohlstand:
> In old comment threads you have said about emulator-side removal side multichip, that it was a barrier for using in OPNMIDI.

One another thing about of multi-chip removal from their side: there are made library to be UGLY: the libOPNMIDI is designed to be used multiple times in same application to generate concurrent streams of different music tracks. There are have put the dozen of objects statically. This is VERY BAD practice even out of the emulator but in the codding commonly. Their library no more allows parallel multiple usage, even don't using it for multi-chip, but using for different instances there are may work in different threads. In result is the slow, clunky, and unsafe solution. Even out of multi-threading, their library allows only one instance which is always built-in and gives no way to use more than one instance in the same app.

Wohlstand:
Anyway, it's will be much easier to backport their changes to our as I have made lot of own changes on original MAME core which in parallel also got some changes.

papiezak:
@jpcima yeah, I meant this version. I just thought it matches official MAME repo, my bad.
P.S.:
>6 channels is too small count of channels to play complex music on the single chip.

If we're talking about complex **MIDI** music, then yes :laughing: Some FM musicians could do magic with MML, but even then, good synth should have at least 28 channels (like SC-55mk2)

Wohlstand:
> If we're talking about complex MIDI music, then yes laughing Some FM musicians could do magic with MML, but even then, good synth should have at least 28 channels (like SC-55mk2)

OPNMIDI is designed to be **MIDI** playing library. And it's a big difference when you are playing music directly on the chip by tweaking it's binary commands, or you are making the MIDI synthesizer that interprets MIDI commands and automatically manages notes between of the chip channels.

jpcima:
@Wohlstand One point of annoyance about merging the changes, is that Genesis-Plus-GX seems to have forked from an older MAME than we have, and the changeset does not let itself apply simply onto the new sources.
So, I'd rather be tempted to add this Genesis-Plus-GX flavor of MAME as another emulator choice, and to reapply the same changeset done to MAME on top of it.

Wohlstand:
It's not hard to directly import their changes directly as MAME wasn't changed so much since the version forked by Genesis-Plus-GX, and possibly some of changes are duplicating.

Wohlstand:
I don't want to directly use `Genesis-Plus-GX`'s version even when we are will repair the nuked multi-chip support by many possible or existing reasons. However, I agree to have it as one another emulator option. I.e. Like I have made on ADLMIDI with Nuked 1.8 and Nuked 1.7 which was optimized by @troosh and it still works a bit faster than the Nuked 1.8 which is still be not optimized. At first it's will be needed to refactor the names to avoid possible linking conflicts (the good one Qt Creator does that easily and well).

ekeeke:
@Wohlstand : the thing is, I am not part of the MAME team and the version of MAME YM2612 core I use in my own repository is not meant to be the official version or developed/maintained with the goal to be a library for outside projects. I removed multi-chip support in my version because it was not needed by my Genesis Plus GX emulator, just like I removed support for other OPN chips that  the original version. You are obviously free to either backport my latest changes in MAME official version or copy my own version and I and multi-chip support back yourself but you can understand I will obviously not add back something I have zero need for in my own project.

jpcima:
I'm not in love with this choice of design even in context, but it does not matter very much honestly.
I's easily done to add back the multiple instance support, and we needn't share strictly identical versions of this code. The OPNMIDI variant is going to diverge anyway, and it needs to have some specific code added to serve for its own purposes.

PR #117, opened 2018-06-20
basic framework of sysex handling and sequencer support

jpcima:
An implementation of sysex with API which was integrated in the sequencer, and tested on real MIDI files.
It's only the framework, no implementation of message themselves, athough it has recognition for few basic ones.


Wohlstand:
I have merged your work into my another branch `split-midi-sequencer-and-sysex` which a mix of my another experimental branch `split-midi-sequencer`. Work is seems completed, however, I polishing it around and then I'll merge it into master.

PR #119, opened 2018-06-22
Split midi sequencer and sysex 2

jpcima:
Handle resets and master volume.
Master volume handles MSB only for 7bit resolution. It was for a quick solution because it makes formulas simpler to adapt.


Wohlstand:
About of Master-Volume, it would be a simpler formula, let's I'll try some experiment on this...

jpcima:
Yes this can be simplified. I've not bothered to make a good fit for now, or create a formula of higher resolution. This was the quick solution.

jpcima:
Here's formula details below which I will commit

```
x=linspace(0, 127^4)
y=8*log2(x/(127^4))+63.49999
# fit y=a*log(x)+b
a=11.541560327111707
b=-1.601379199767093e+02
```

PR #120, opened 2018-06-23
glide: handle for all the midi channels

jpcima:
Doing the iteration on channels the same as other code does, because the MIDI channels seem not fixed at 16.


Wohlstand:
Yeah, when some MIDI files will have multiple "devices" are will extend MIDI channels count.

PR #124, opened 2018-07-02
color attribute for channel descriptor

jpcima:
This puts a color ID which identifies the MIDI channel in low 4 bits of the channel description.
This extends API to allow implementing a colored output.
(adljack currently has a monochrome output of channels)

Upper bits are reserved for keeping more infos as necessary. (key release was discussed before)

### DIFF
--- a/src/adlmidi_midiplay.cpp
+++ b/src/adlmidi_midiplay.cpp
@@ -1761,7 +1761,11 @@ void MIDIplay::describeChannels(char *str, char *attr, size_t size)
             }
         }
 
-        attr[index] = '\0';  // TODO
+        uint8_t attribute = 0;
+        if (loc)  // 4-bit color index of MIDI channel
+            attribute |= (uint8_t)(loc->loc.MidCh & 0xF);
+
+        attr[index] = (char)attribute;
     }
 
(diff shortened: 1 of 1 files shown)


Wohlstand:
One note: while on UNIX-like systems are Escape sequences are used which makes colorization easy job, on Windows it sucks strongly (and also, you will need to map colors as UNIX-side (both Linux and macOS) and Windows are has different maps of all 16 console colors), and colorization requires you every time call the special WinAPI function to change the console output.

jpcima:
I will not use the UNIX escape codes.
For me, it will be the usage of 256 color extension of ncurses on xterm, see [note](https://github.com/jpcima/adljack#why-are-the-colors-odd-looking-in-a-x11-terminal).

For Windows, it will be the Pdcurses terminal window, not Windows console, so no such issue will present itself in that regard.

Wohlstand:
Nice! :fox_face: :smiley:

jpcima:
I find red-colored channel 0 to be suspiciously over-present in most of songs.
It really raises for me some suspicion on the validity of the value `MidCh` kept in `Location` structures..

Issue #125, opened 2018-07-03
Add the multi-bank map over embedded banks


Wohlstand:
Provide an ability to construct dynamical multi-bank from multiple embedded banks.

The suggested API for that (may be changed if needed):

```c
typedef struct 
{
    /*! Use percussion set, otherwise, melodic set will be used */
    int is_percussion;
    /*! ID of embedded bank to use */
    int embedded_bank_id;
    /*!  MSB MIDI bank number */
    ADL_UInt8  midi_bank_msb;
    /*!  LSB MIDI bank number */
    ADL_UInt8  midi_bank_lsb;
} ADL_BankMapEntry;

/* Constructs the dynamical multi-bank from off the set of embedded banks. 
Returns 0 if everything is fine, otherwise -1 when any error has occurred.
For example, given invalid input data or missing some of required data. */
int adl_setBankMap(struct ADL_MIDIPlayer *device, const ADL_BankMapEntry *bank_map, int bank_map_size);
```

Requested by @kode54

jpcima:
It's a redundant identification of the bank, as there is already the structure `ADL_BankId` to identify.
I can also suggest a function to complete the set of bank API which exist already.

```
ADL_BankId id = [...];

// get/create, and obtain slot iterator
ADL_Bank bank;
adl_getBank(device, &id, ADLMIDI_Bank_Create, &bank);

// load embedded bank into slot
int bankIndex = 0;
adl_loadEmbeddedBank(device, &bank, bankIndex);
```

Wohlstand:
Can this assign different melodic and drumsets from different banks?

jpcima:
Yes: `ADL_BankId` has the `percussion` flag. Depending of this flag, the code can take lower or upper 128 instruments of the embedded bank.

Wohlstand:
Made! Tested, and it works! (I have done the experiment on adlmidiplay side with using of those functions, and result works fine!)

PR #126, opened 2018-07-03
embedded bank loading API into multi-bank

jpcima:
Proposal to implement like suggested in #125. Not tried


Wohlstand:
When embedded banks are disabled, you must to set play's error that this thing built without of embedded banks.

Wohlstand:
Okay, because App-Get failed, I restarted the build, it must be fine now...
Also, I'll try to test this thing in work, it's very curious how it will work...

jpcima:
Semaphore says it can have a cache of debian packages. (to use in place of apt-get)
https://semaphoreci.com/docs/how-to-install-dependency.html#caching-installations

Wohlstand:
Oh, that would help a lot!

Issue #30, opened 2017-11-10
[XMI 2 MIDI] Implement track spliting


Wohlstand:
Because current converter does merge of all songs inside same XMI file, result becoming too cluncky. As example, music from "System Shock" game is widely uses multi-song XMI files.

XMI2MIDI: https://github.com/Wohlstand/libADLMIDI/blob/master/src/cvt_xmi2mid.hpp
* [x] Provide a way to choice XMI track number file open option (`adl_open[FileData]Track()` which will have one extra field - a track number). By default, open first XMI track.
* [x] Inside of XMI2MIDI converter, ignore data of unneeded tracks.

Interrupt:
This would be great to have, as an experiment I've tried using ADLMIDI to play the XMI music in our System Shock source port but with all the tracks playing at once it's just a cacophony!

WIP here: https://github.com/Interrupt/systemshock/tree/adlmidi

jpcima:
This is an API proposal to control which of the tracks can be played by the sequencer.
It has benefit of not being tied to XMI particularly, so it can work with the rest of formats.

```
enum ADLMIDI_TrackOptions
{
   ADL_TrackOption_On,
   ADL_TrackOption_Off,
   ADL_TrackOption_Solo,
};

/*Count tracks of current data in sequencer*/
int adl_getTrackCount(struct ADL_MIDIPlayer *device);

/*Set track options in sequencer*/
int adl_setTrackOptions(struct ADL_MIDIPlayer *device, int trackNumber, unsigned trackOptions);
```

On System Shock side:
`adl_setTrackOptions(adlDevice, themeID, ADL_TrackOption_Solo)`

Interrupt:
@jpcima This works pretty well! It has some issues with SS1's songs though, like if I play track 10 from `thm1.xmi` in Solo mode it sounds like a good melody, but if I turn off all of the tracks and just enable that one it sounds super strange.

jpcima:
@Interrupt you may want to try an update considering @Wohlstand has worked hard to implement a support of XMI loop events.
I made some experiments in your `adlmidi-xmi` branch (before a Music AI commit), and I was able to hear some properly looped music that wasn't before.

I noted some things not supported yet, which I don't know if needed or not.

- [ ] CC 120 branch, and RBRN block (found in `thmintro`, `introthm`)
- [ ] CC 119 callback trigger

Interrupt:
[...]

PR #133, opened 2018-07-09
xmi: handle AIL callback trigger

jpcima:
#30 

Implement AIL callback trigger (`thm0.xmi` first track)

I make this controller event a `ST_SPECIAL` in the sequencer, as done for loop events before, and covered under the XMI file format case. But if prefered, it could be left in controller form.

This is modeled after AIL functions `AIL_install_callback`, `AIL_cancel_callback`, except I implement as one function. There is one callback to handle all the triggers, and the trigger value (CC value) is received by the callback.

### DIFF
--- a/src/midi_sequencer.hpp
+++ b/src/midi_sequencer.hpp
@@ -131,6 +131,8 @@ class BW_MidiSequencer
             ST_LOOPSTACK_END   = 0xE5,//size == 0 <CUSTOM>
             //! [Non-Standard] Loop End point with support of multi-loops
             ST_LOOPSTACK_BREAK = 0xE6,//size == 0 <CUSTOM>
+            //! [Non-Standard] Callback Trigger
+            ST_CALLBACK_TRIGGER = 0xE7,//size == 1 <CUSTOM>
         };
         //! Main type of event
         uint8_t type;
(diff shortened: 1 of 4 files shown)


src/midi_sequencer.hpp
```diff
@@ -453,6 +455,19 @@ class BW_MidiSequencer
...
+     * @param userData Pointer to user data (usually, context of something)
+     * @param trigger Value of the event which triggered this callback.
+     * @param track Identifier of the track which triggered this callback.
+     */
+    typedef void (*TriggerHandler)(void *userData, unsigned trigger, size_t track);
+
+    //! Handler of callback trigger events
+    TriggerHandler m_triggerHandler;
+    //! User data of callback trigger events
+    void *m_triggerUserData;
```
Wohlstand: You forgot to `NULL` them on constructor initialization!

src/midi_sequencer_impl.hpp
```diff
@@ -1518,6 +1531,17 @@ void BW_MidiSequencer::handleEvent(size_t track, const BW_MidiSequencer::MidiEve
             }
         }
 
+        if(evtype == MidiEvent::ST_CALLBACK_TRIGGER)
+        {
+#if 0 /* Print all callback triggers events */
+            if(m_interface->onDebugMessage)
+                m_interface->onDebugMessage(m_interface->onDebugMessage_userData, "Callback Trigger: %02X", evt.data[0]);
+#endif
+            if(m_triggerHandler)
```
Wohlstand: That will crash in release build :confused:

PR #138, opened 2018-07-18
fix the conversion of FMIns->ADLI operators

jpcima:
Rewrite the conversion behind `adl_getInstrument` which is incorrect
The operators don't have correct values.
In VST programming, I have found this when the bank management doesn't show the same values as OPL3BankEditor. I did a quick check after the patch and operators seems to match.
(with some exceptions as Level, KSR, but because the bank editor has these values inverted)


jpcima:
Yeah I've myself decided to implement the same inversion of values for the UI.
I've made the code changes which display values now identically to OPL3BE.

PR #141, opened 2018-07-22
implement the velocity offset parameter

### LINKED ISSUE #140: MIDI velocity offset not implemented
There is this parameter from WOPL files which doesn't have an implementation in libADLMIDI.

It's also not stored in adl meta-instruments, and should receive support in `adlmidi_load` for conversion in both directions.

jpcima:
Fixes #140 

This adds to velocity on a note on event and keeping the result value in 1-127 range.

It has the conversion between the internal and external patch format.
For embedded banks they don't have it, in that case it's a zero value.


Wohlstand:
The parameter came from Apogee Sound System where it is specified in the TMB file format. Yeah, it's need to extend gen_adldata to take it from TMB and WOPN files, and write into embedded banks database

Wohlstand:
Since I have added the "real 4op" flag to allow 4op instruments be correctly work when two equal  `adlins`s, so, It's no more need to keep equality of this.

Wohlstand:
As now I'm very tired, I'll take this to extend tomorrow... I gonna to equip GenAdldata with support of velocity offset value for bank formats are supports it (TMB and WOPL).

jpcima:
I've already pushed some commits adding the support in GenAdldata for the 2 formats.
You can diff the adldata.cpp and see.

Wohlstand:
Just for curious, I have checked which instruments in Duke Nukem 3D bank are having velocity offset, and have tried to compare libADLMIDI's and PM.exe's output, and seems there are similar... Going to capture WAVs from both and compare them...

jpcima:
This is a recording of the track after I've applied the fix of velocity-offset.
Also, this has permitted me to build the support in VST and it can serve to experiment manipulating velocity offset in real time.

[d3d-ADLMIDI-2.zip](https://github.com/Wohlstand/libADLMIDI/files/2230349/d3d-ADLMIDI-2.zip)
[d3d-PM.zip](https://github.com/Wohlstand/libADLMIDI/files/2230350/d3d-PM.zip)

PR #142, opened 2018-07-23
preserve the data of operator 3-4 in internal storage, regardless of being 2op/4op

jpcima:
Following your comment on 4op storage

This allows to keep operators 3-4 in program structures which are 2op.
This also eliminates the logic code which uses adlins comparisons for 2op checking and replaces with usage of flags.

I did a quick test.

### DIFF
--- a/src/adldata.hh
+++ b/src/adldata.hh
@@ -49,6 +49,9 @@ struct adlinsdata
 {
     enum { Flag_Pseudo4op = 0x01, Flag_NoSound = 0x02, Flag_Real4op = 0x04 };
 
+    enum { Flag_RM_BassDrum  = 0x08, Flag_RM_Snare = 0x10, Flag_RM_TomTom = 0x18,
+           Flag_RM_Cymbal = 0x20, Flag_RM_HiHat = 0x28, Mask_RhythmMode = 0x38 };
+
     uint16_t    adlno1, adlno2;
     uint8_t     tone;
     uint8_t     flags;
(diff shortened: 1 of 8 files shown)


Wohlstand:
I have ran quick test on DMXOPL3-GS's bank (which does wide usage of Pseudo-4op instruments) and found some dirty sounding with the patch...
[Natols (Zero Wing) (arranged).mid.zip](https://github.com/Wohlstand/libADLMIDI/files/2221402/Natols.Zero.Wing.arranged.mid.zip)
This set reproduces bug.
on Master sound is fine

jpcima:
If there is such a problem, it is possibly in the bank loading because that was a part not covered in my quick test, as my synth had a custom implementation of bank loader.

So first, I wrote a unit test to check validity of conversions back and forth with WOPL files.
It has revealed only the rhythm mode flags not being handled by conversions, so I implemented them in that commit. Also, I took conversion functions out of their source file for access from unit tests.

Gotta check now this Zero Wing music.

jpcima:
These midi and bank sound good to me in midiplay.
Are you sure to have latest, and that your cached build has not played you a trick?
The first commit had received a forced push from me for a bugfix, not too long after.

Wohlstand:
What the sense of a bug:
while playing sound, there are some junky notes (sounds are played incorrectly) are can appear in the process of playing, that doesn't happens on begin, but in the process

jpcima:
I have fixed the problem. Now, a same register dump is produced as in the master branch.
Apparently, it was important that `Phys::voices[]` contain the same data when it's in the 2op mode.

Wohlstand:
Looks much better now! :fox_face: :+1: 
However, I think, it's need to improve it's flags system to don't require copy-pasta here...

PR #146, opened 2018-07-30
hide private symbols

jpcima:
This makes the internal symbols of the library hidden.

It's an attempt for me to fix the case of having libADLMIDI and a copy of Measurer in the same program without clashes of symbols.
When kode54 made the latest changes, importing them in VST has crashed it, because both my program and library have Dosbox OPL, but now they have differed and the structure layout was changed.
(it suggest my program was picking the Dosbox OPL of libADLMIDI and has smashed my stack because of this)

~~In case of GCC, it forces the static library to emit only the public symbols. (needs CMake 3.3+, hence a bump of version)
On MSVC and others I don't know what can happen: there is not an equivalent to my knowledge of visibility attributes (it's only for DLL). Maybe you know better what to do about this.~~


jpcima:
Other strategy, in a way which is compatible with cmake 3.2.
GCC and Clang support only

Wohlstand:
The MSVC case can be kept as-is, that can be done later :fox_face: :wink:

jpcima:
I rename the symbols in my plugin, to make sure a MSVC user is not going face problems with the link of static libraries.
The more important thing is about cleaning up the interface of the linux shared library.
In a little bit, I'll do the same of libOPNMIDI.

PS. there's a few of CMAKE_COMPILER_IS_GNU checks. they don't cover the detection of clang compilers.

Issue #154, opened 2018-08-26
Intempestive arpeggio after long pause times


jpcima:
I made a reporting of arpeggio which I find to happen, on long running instances of a synth with libADLMIDI. (in this case, many hours)

Here is a particular report after starting a MIDI which used piano only on DMXOPL.
On a fresh instance of synthesizer, there aren't occurrences of any arpeggios.

The reports are taken at the start of `prepareChipChannelForNewNote`. (ie. just after selection, before channel allocation proper)
http://paste.alacon.org/45400

The context was 8 chips, 8 four-ops and pseudo-4op instrument.
It shows the selection of chip channel made and a current context.
The scores have [m] incompatible channel category, [x] occupied by 4op primary.

As shown, the channel selector is setting the preference on a channel which was already occupied, but had for it the largest of `koff` times, which dominates the scoring.

I suspect it's possible that a channel does not have an updated `koff` at a time when it's evaluated.
The updated of channels' ages are performed after audio cycle in `TickIterators`. But there is not necessary an audio cycle in between two arrivals of note-on events.

![arpeggios](https://user-images.githubusercontent.com/17614485/44632614-0cc4d600-a97e-11e8-9b92-6b6e87441e07.png)

jpcima:
It's been fixed.

PR #155, opened 2018-08-27
early koff update to prevent a bad reuse of channel

jpcima:
#154 
I have thought of a simple attempt to fix arpeggio issues. It's not tested. :warning: 

It would force an update of timing just at the end of note-on, and before a second note-on which would arrive without having `TickIterators` happened between.
In such a way, it's become possible that a taken channel can reinitialize its `koff` time before a case of note-on coming right after.
In such a way, it could fix the scoring of a next channel selection, which could happen to reselect the already taken channel if the `koff` time outweighs the 4000 point malus for each arpeggio user already on channel.

Do you think of it as correct reasoning?

### DIFF
--- a/src/adlmidi_midiplay.cpp
+++ b/src/adlmidi_midiplay.cpp
@@ -578,6 +578,15 @@ bool MIDIplay::realTime_NoteOn(uint8_t channel, uint8_t note, uint8_t velocity)
     }
 
     noteUpdate(channel, ir.first, Upd_All | Upd_Patch);
+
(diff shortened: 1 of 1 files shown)


jpcima:
As I've given myself this patch some testing, I've observed so far:
none of violent arpeggio which is usually after long idle periods can be heard!

(For ordinary case arpeggios, those happen way rarely but in related ways with goodness formula.)

Attaching the arpeggio reporting patch (govnocode for test purpose only)
[adlmidi-reporting.diff.gz](https://github.com/Wohlstand/libADLMIDI/files/2325343/adlmidi-reporting.diff.gz)

PR #163, opened 2018-08-30
fix a case of miscalculation and infinite loop

jpcima:
Important bug fix :warning: 

This solves a problem of execution getting stuck in sound processing.
In following vibrato depth calculation, the value is typed as unsigned, and creates an expression result as a very large integer. This large integer gets into the frequency calculation formula and creates an infinity result. It puts the program in a situation of infinite loop at OPL3::noteOn.


Wohlstand:
Oops!

PR #164, opened 2018-08-31
gen_adldata: conversion of AdLib Gold BNK2 as 4op

jpcima:
This is probably the good fix to enable the BNK2 loader to output the 4op instruments.
In current adldata, the Master of Magic "4op" has only a 2op instruments set.


Wohlstand:
Damn! Thanks for a fix!

PR #165, opened 2018-08-31
fix a miswritten name: sneakernets

jpcima:
Found another small error in metadata.
This one, it's just in prefix which I don't know what its usage is. Adldata seems unaffected.


Wohlstand:
Lol! :rofl: Why I typed "Skeakernets"? Maybe "sKeakerKets"? xD
Okay, thanks for a fix :fox_face: :+1:

PR #166, opened 2018-09-01
provide default initialization for bank setup

jpcima:
Fixes an uninitialized use which is difficult to track.

With embedded bank disabled, the bank setup never initializes the value at `adl_init`, which depends on values for channels setup as rhythm mode or not, and also modulator scaling to invoke codes which silence all notes.


Wohlstand:
Woops!
Thanks for a fix!

jpcima:
Thanks. MIDI timing is working now on biggest of latencies, to let you know.
I have used some time to focus bit on more minor, side things.

PR #171, opened 2018-09-06
access into global chip settings

jpcima:
Add some functions to be able to track chip's global state which is current: tremolo flags, volume model.


src/adlmidi.cpp
```diff
@@ -403,6 +403,13 @@ ADLMIDI_EXPORT void adl_setHVibrato(ADL_MIDIPlayer *device, int hvibro)
     play->m_synth.commitDeepFlags();
 }
 
+ADLMIDI_DECLSPEC int adl_getHVibrato(struct ADL_MIDIPlayer *device)
```
Wohlstand: Oops! :yum:
You must to use `ADLMIDI_EXPORT` here.

Issue #172, opened 2018-09-10
adl_metaMarker doesn't properly check pointers


GabrielRavier:
It uses the device and the play pointer (obtained from calling GET_MIDI_PLAYER with device) without checking for them to be nullptr, and then checks them, which is either useless or a total error

(See adlmidi.cpp, line 911-917)

jpcima:
Thanks for reporting. To add, I think the null checking would need some clean up globally.
In the case of this function, null is unchecked at the entry and has a contradiction, as you have observed.

The implementation has numerous inconsistencies, where it's sometimes both of `device` and `play` checked, only `device`, or no check (example `adl_setBank`).
As we can normalize the way of doing checking, I would think it's the good idea to keep it at one null check, possibly to `assert` the second, considering one can't be not-null without the other.

Wohlstand:
Oh, this is a missing!
Yeah, it's need to check for null both the `device` and `play`... Yeah, assert can be used in `play`.

Wohlstand:
@GabrielRavier , Please review and test the @jpcima 's stuff, I hope it's fine now :fox_face: :wink:

GabrielRavier:
> 
> 
> Closing as this was fixed, but no reply was received from @GabrielRavier to confirm that fix is fine.

I'm bad at looking at my emails 😛

GabrielRavier:
Btw the fix is fine

PR #173, opened 2018-09-11
have consistent null checks and returns

jpcima:
#172 

Harmonize the way how are performed the null checks in the public functions, and some other things.

### DIFF
--- a/src/adlmidi.cpp
+++ b/src/adlmidi.cpp
@@ -60,7 +60,7 @@ ADLMIDI_EXPORT struct ADL_MIDIPlayer *adl_init(long sample_rate)
         return NULL;
     }
 
-    MIDIplay *player = new MIDIplay(static_cast<unsigned long>(sample_rate));
+    MIDIplay *player = new(std::nothrow) MIDIplay(static_cast<unsigned long>(sample_rate));
     if(!player)
     {
         free(midi_device);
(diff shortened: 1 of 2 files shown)


src/adlmidi.cpp
```diff
@@ -60,7 +60,7 @@ ADLMIDI_EXPORT struct ADL_MIDIPlayer *adl_init(long sample_rate)
         return NULL;
     }
 
-    MIDIplay *player = new MIDIplay(static_cast<unsigned long>(sample_rate));
+    MIDIplay *player = new(std::nothrow) MIDIplay(static_cast<unsigned long>(sample_rate));
```
Wohlstand: Will this `(std::nothrow)` compatible to C++98?

jpcima: It doesn't have a (C++11) mark on cpp reference so I assume, yes.
It has built on my djgpp copy. Now that remains to check if Watcom will accept.

Wohlstand: The bottleneck for now is an official Android NDK where the C++98 mode is default one.

Wohlstand: Oh, what if I'll add the build check on Android NDK too?

Wohlstand: Okay, I'll merge the stuff, but later I'll make the CI build for Android to verify the build on it also. It's will be better than I'll do this manually which I doing update of my Android apps not so offten.

jpcima: I could as well have kept `new` the way like it is, but remove another null check down below.
Having new and a null check together just don't make sense.

Issue #178, opened 2018-09-24
Optimize initialization of DosBox emulator when using chip native sample rate


Wohlstand:
While Nuked OPL3 does almost instant initialization, DosBox while initialization, does generation of heavy arrays.

It's need to pre-generate them for chip native sample rate, and put them as a separated file. When using this rate, do the silly memcpy of that data rather heavy computing. Use dynamic generation for the case where requested rate doesn't matching chip native.

The function to optimize:
https://github.com/Wohlstand/libADLMIDI/blob/master/src/chips/dosbox/dbopl.cpp#L1297

jpcima:
I don't mind. A generated array makes a smaller binary.
How impacting is this problem on a low performance system?

Wohlstand:
On my phone when I opening any music file and chips are doing initialization, the waiting pause stucks the application. That also happens on powerful systems: in the VLC player a little pause on an attempt to play any song.

jpcima:
Let me propose another thing: cache these tables for the sampling rates used, and share them across instances. Done this way, instantiating the first time will be slow but will be fast next times.

Wohlstand:
Anyway, let's keep one (chip native) as the default state, and when it's need to change sample rate, then, re-generate into something another

jpcima:
You have to consider multi-instances for different sample rates.
I think that the computed data is best kept in a table of tables associated by sample rates, to keep protected by read-write-lock. Native chip rate needn't have a particular case made for it.

Wohlstand:
I may guess, the current one will fail when having multiple instances with different sample rates as tables you can see are static globals.

jpcima:
The initialization should be like this.

```
take the reader's lock;
if (my rate is in map)
   take that table pointer as mine;
else {
  take the writer's lock;
  compute the table for my rate;
  set it as mine;
  insert it in map;
}
```

Wohlstand:
Mission completed!

My Android build of demo app to test yourself:
[ADLMIDI-Player-release.apk.zip](https://github.com/Wohlstand/libADLMIDI/files/2413049/ADLMIDI-Player-release.apk.zip)

(it sucks because it doesn't using foreground service, therefore it can suddenly die in background, and I working on a fix of this - learning foreground services on Android)

PR #179, opened 2018-09-24
dosbox: add caching of the attack rate computation

jpcima:
#178 

A quick-made modification to DOSBox emulation permitting to cache the table computations which depend on sample rate.
Please check.


Wohlstand:
Looks like it's fine! :fox_face: :+1: 
In my VLC it's no more waits until start playing the song, gonna to test the result on my phone...

jpcima:
Yeah. only the mutex wrapper is kinda crap looking, I've tried keeping it good for c++98.

PR #180, opened 2018-09-25
dosbox: fix a potential cache access problem

jpcima:
For some reason I rethought of last Dosbox changes, and came the "damn" moment about a mistake made.
As the cache entry is referred inside its vector container and the lock is released, it could be that another pushes to the vector near the same moment, invalidating the item's reference. It can do an invalid access after the line 1461.

Not a likely problem for sure, here is the fix.

### DIFF
--- a/src/chips/dosbox/dbopl.cpp
+++ b/src/chips/dosbox/dbopl.cpp
@@ -38,6 +38,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <vector>
+#include <memory>
 #include "dbopl.h"
 
 #if defined(__GNUC__) && __GNUC__ > 3
@@ -80,7 +81,8 @@ private:
 #if !defined(_WIN32)
 #include <pthread.h>
 struct Mutex : NoCopy {
-	Mutex() : m(PTHREAD_MUTEX_INITIALIZER) {}
+	Mutex() { pthread_mutex_init(&m, NULL);}
+	~Mutex() { pthread_mutex_destroy(&m); }
 	void lock() { pthread_mutex_lock(&m); }
 	void unlock() { pthread_mutex_unlock(&m); }
 	pthread_mutex_t m;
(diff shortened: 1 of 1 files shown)


Wohlstand:
Huh?

jpcima:
I made a mistake a mistake in the "fix". lol wait

jpcima:
The thing is, it would need entries allocated by new, so the vector resized will not break references.
Then, I would like to have entries as RAII, so the library can load and unload clean. (plugin use)

Problem, c++ 98 does not have `shared_ptr`. Is `std::tr1::shared_ptr` ok to use?

Wohlstand:
Then, you need to use hand-made one, like I have used on libADLMIDI side. This thing would be some sort of simplified but keep same RAII concept.

jpcima:
I think of the current fix as fine, and fixed a warning along the way about the pthread mutex.
Built the file on gcc and clang with -Wall, both c++ 98 and 11 configurations.

Wohlstand:
Then it's need to validate the build on official Android NDK to be sure this will work. The reason I don't use std::*ptr that Android NDK have lacked them.

jpcima:
Oh I can check it. Which version?

Wohlstand:
Primary, latest

jpcima:
I picked ndk r18 with ARM. It built in both of c++98 or c++11, with either kind of ptr.
About this "arm 4.9" toolchain, I noted it contains a clang wrapper in place of g++, and has the STL substituted with llvm's.

`arm-linux-androideabi/bin/clang++ -std=c++11 -c dbopl.cpp`
`arm-linux-androideabi/bin/clang++ -std=c++98 -c dbopl.cpp`

Wohlstand:
I hope, this will be built fine with any ABI, as my phone does using 32-bit ARM eabi-v7, let's check that...

jpcima:
The ndk-16b arm platform was able to compile this based on gcc.
The latest had a problem starting a build of your player, because it lacks the mips64 target, removed from r17.

Wohlstand:
Me at home now, which means, I'm able to test the stuff on a place now! :fox_face: :+1:

Issue #6, opened 2017-02-22
Implement SIMD for Nuked OPL3 Synthesizer


Wohlstand:
This is a very accurate OPL3 emulator, but it requires more powerful CPU than DosBox's emulator. To improve performance of this, I have to try implement SIMD support for it. (both SIMD and non-SIMD versions are will be kept to be able compile for CPUs without SIMD support).

troosh:
My timid steps in this direction: https://github.com/adplug/adplug/pull/62

PR #183, opened 2018-10-04
improvement on the installation procedure based on CMake

jpcima:
- Detect the standard installation paths of OS distributions.

A fedora guy has hinted me about needing this to have the libraries in correct places where the distributions intend them to be, especially in some multi-lib layouts.
One can override this by `CMAKE_INSTALL_*DIR`.

- Install a version symlink for the shared library.

This will let a linker use a version tagged library identifier.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,8 @@
 cmake_minimum_required (VERSION 3.2)
 project (libADLMIDI C CXX)
 
+include(GNUInstallDirs)
+
 #===========================================================================================
 # Strip garbage
 if(APPLE)
@@ -306,6 +308,7 @@ if(libADLMIDI_SHARED)
     if(WITH_EMBEDDED_BANKS AND WITH_GENADLDATA)
         add_dependencies(ADLMIDI_shared gen-adldata-run)
     endif()
+    set_target_properties(ADLMIDI_shared PROPERTIES SOVERSION "1")
 endif()
 
 if(NOT libADLMIDI_STATIC AND NOT libADLMIDI_SHARED)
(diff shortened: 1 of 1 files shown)


Wohlstand:
Looks I need to apply same to dozen of my other libs include AudioCodecs and SDL Mixer X...

jpcima:
Have lots of fun! gotta repeat on OPNMIDI side as usual. I do such changes as I prepare to submit a new archlinux package release.

I will publish as shared library as I think it will need it for addition into the descent game.
It's not possible to link it right in because of incompatibility to the GPL. I've suggested a possibility of runtime link which I think is the adequate solution. Feel free to add on [issue 407](https://github.com/dxx-rebirth/dxx-rebirth/issues) if I said any wrong here.

Wohlstand:
About of license: I wanna to distribute library as LGPL completely, but to make that I need to confirm @bisqwit to license parts are originally written by him as LGPL than GPL. For the case of DOSBox emulator which is GPL-licensed, it's still be optional in the build as Nuked cores are in use, and both of them are LGPL as nukeykt have granted that.

jpcima:
The documentation inside of ADLMIDI indicates GPL3. Have you contacted Joel asking if he would be interested to make a licensing exception?
I thought about Nuked it may be worth a study whether it can be made into a kind of Nuked-"lite" for economy of computation, such that we may have an alternative to GPL Dosbox.

Wohlstand:
Related to Nuked, I have this issue: https://github.com/Wohlstand/libADLMIDI/issues/6
Mainly to optimize Nuked as possible

bisqwit:
Currently ADLMIDI does not contain a component called “libADLMIDI”. However, I am interested about that prospect, and I would be open to licensing that kind of subcomponent — which I repeat, does not currently exist at least in my repository — under LGPL.

PR #184, opened 2018-10-05
reorganize the OPL3 chip manager code

jpcima:
I move the definitions of the OPL3 chip manager in a file of its own.
Next I intend to do the same of `MIDIplay`.

Tested as building: Linux GCC, DOS DJGPP, unit tests


jpcima:
Moved `MIDIplay` to its own header.
While doing this, I modified some members of `MIDIplay` such that their contents are dynamically allocated and the structure keeps a unique smart pointer on them.
The first benefit is that this fat structure is going to be made smaller, with some potential benefit with regards to caching. The second thing is about forwarding their types to minimize coupling of header files.

jpcima:
Next, I've modified `adlmidi_private.hpp` so that it will be much much lighter, by instead of including the full headers of everything, it will only declare forward declaration as much as possible. I make it remain a centralized place where are included most the standard definitions and the compiler workarounds on these.
By this way, I allow the `midiplay` and `opl3` parts to include this from their own headers.

src/adlmidi.cpp
```diff
@@ -112,13 +120,14 @@ ADLMIDI_EXPORT int adl_setNumChips(ADL_MIDIPlayer *device, int numChips)
     else if(play->m_setup.numFourOps < -1)
         play->m_setup.numFourOps = -1;
 
-    if(!play->m_synth.setupLocked())
+    OPL3 &synth = *play->m_synth;
```
Wohlstand: Please make a typedef to unify with OPNMIDI when backporting changes, like I made with MIDI Player class :wink:

jpcima: There doesn't exist any currently, is this correct? I have to make it?

Wohlstand: I wasn't made any typedefs for Synth as I never used it's name in this file, but that you have made just now is good now :fox_face: :+1:

Issue #186, opened 2018-10-19
Provide pkg-config support


Wohlstand:
Pkg config files are helps various build systems to recognize dependencies and their setup to easier deal with them.

What's need to make:
* adlmidi.pkg.in where CMake is able to pass version number and flags for linking. 
* Let CMake script generate the final file and install it into the system.
* similar for OPNMIDI.

jpcima:
I made the pkg-config. Pushed into this work branch on here.
https://github.com/Wohlstand/libADLMIDI/tree/pkg-config

(tried making a PR, met a temporary error of Github. PR isn't here and it won't let me redo.
Never seen such error before. Probably broken for the time being, hence why doing this)

Wohlstand:
Nice! :3
I'll test this out, and, looks, like I need to make the version to be in one place (for now, it's the part of the public header file).

Wohlstand:
I think, it's fine now,, and no more need to keep this opened. Any further issues will be discussed separately.

PR #189, opened 2018-10-24
cmake rule to fail the build on version mismatch

jpcima:
To complete pkg-config #186 

This adds a rule to compile a simple check program, in such a way that it's forbidden to forget setting the version number. Except if the developer works only with QMake not CMake, but then it will fail the CI.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -590,6 +590,16 @@ if(WITH_UNIT_TESTS)
     add_subdirectory(test)
 endif()
 
+# === Version check ====
+set(VERSION_CHECK_SOURCE "
+    #include \"adlmidi.h\"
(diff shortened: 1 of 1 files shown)


Wohlstand:
Nice! Yeah, it's a good thing to avoid cases when you forgot to change the version in both header and the CMake script! :fox_face: 
I'll check this again a bit later and will merge.

Wohlstand:
I think, you are able to make the compile test without or making a separate task, I have used this to check support of sse/neon at Audio Codecs / libOpus where I have made own CMake build where it hadn't it.

jpcima:
It's different cases. The SSE check is just one time.
This checks also when you edit the header so it will fail on next make.

PR #191, opened 2018-11-10
fix minute display on 1+ hour tracks

jpcima:
Add a call `fmod` to fix the minutes. Also remove the `floor` calls redundant because of casts.

MIDI file for test [thrust.mid.gz](https://github.com/Wohlstand/libADLMIDI/files/2568453/thrust.mid.gz)

before `11:673:58,221`
after `11:13:58,221`


Wohlstand:
Oops...
But, thanks for a fix! :fox_face:

PR #192, opened 2018-11-10
Std structures

jpcima:
Here is a work in progress for generic internal data structures.

- [X] the preallocated linked-list container `pl_list`
- [X] users
- [X] active notes
- [X] **unit tests** (updated on libOPNMIDI side)

The `pl_list` is a generic container with STL-like behavior, with particularities.
- the `value_type` is the linked-list cell `pl_cell<T>` (it is not `T`)
- the cell consists of a pair of link pointers and value
- the wrapped value is obtained by `mycell->value`
- the list has a sentinel element to mark the end of list.
  1. it's the cell refered by the `end()` iterator
  2. it's the only case of cell which does not contain value
  3. it can be navigated back from it, meaning you can safely do `operator--` on the end iterator
- the end element has `next` to null, so end of list can be checked on this condition `iterator.is_end()`, it's not needed the inconvenience of comparing to `end()`
- (supports external buffering policy. but currently unused)

It's the Boost license tag on these sources of containers I made, because I find these sources valuable for general reuse. (ok to discuss license if you want another)

### DIFF
--- a/src/adlmidi_midiplay.cpp
+++ b/src/adlmidi_midiplay.cpp
@@ -469,7 +469,7 @@ bool MIDIplay::realTime_NoteOn(uint8_t channel, uint8_t note, uint8_t velocity)
     if(isBlankNote)
     {
         // Don't even try to play the blank instrument! But, insert the dummy note.
-        MIDIchannel::notes_iterator i = midiChan.find_or_create_activenote(note);
+        MIDIchannel::notes_iterator i = midiChan.ensure_find_or_create_activenote(note);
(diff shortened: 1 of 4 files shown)


jpcima:
This change has worked pretty robustly on my side for both of plugins.

For some remarks on double-note support which can be made after this:
- instead of "find or create" a note onto the list, it can just push the second. (but it must check for capacity remaining)
- the notes insert at the back; removal takes from the front. it means note on/off would operate as FIFO, the old note keys-off first.

Issue #196, opened 2018-12-12
Cleanup CMake


Flamefire:
The new project structure looks great and the CMake is also mostly fine. However may I suggest the following changes (maybe missing some) to make it cleaner:

- Put targets into subfolders -> `if(OPTIONAL_FEATURE) add_subdirectory(optional_feature) endif()`
- Use ONLY `target_*` directives (as far as possible)
- Remove CMake version checks (some [wrong] checks check for <=3.2 but this is already enforced)
- Remove `add_definitions(-DNDEBUG)` -> CMake does that where required
- `OPENBSD_LOCALBASE` partially ignored -> use it
-  c++11 or c++98? You have differing settings in the CMakeLists. I suggest to remove it and use the compilers default which allows using this in either
- remove `message` on end -> Not "the CMake way". You have set those manually anyway or you have ccmake to easily check those. Better: Rename option to have common prefix, then ccmake/cmake-gui groups them nicely and you don't clash names with super projects
- configure a config header for things like version (https://github.com/Wohlstand/libADLMIDI/blob/d89084d6eaf3bb65c134201f61449656226cbf80/include/adlmidi.h#L3) or options (defines currently passed on command line)

Wohlstand:
> Put targets into subfolders

Yeah, the thing I must to do long time ago, but didn't yet...

> Remove CMake version checks 

Reason why version checks are added to allow support of older CMake builds, but, as required CMake version is higher than in condition, yeah, nuke it.

> OPENBSD_LOCALBASE partially ignored -> use it

Yeah, that needs some more depther research, I did the only small test from off VirtualBox environment...

> c++11 or c++98? You have differing settings in the CMakeLists. I suggest to remove it and use the compilers default which allows using this in either

Most of parts must be enforced with C++98 to be sure code will be built on older compilers (as example, older Android NDKs, OpenWatcom where I have done some experiments, ancient GCCs, and some other clunky compilers I didn't tested yet, but they will more work than not). There are two cases are using C++11:
* `adlmidi2` is enforced to be C++11 app as it's the reincornation of original ADLMIDI player in it's original form where the C++11 was used widely.
[...]

PR #199, opened 2019-01-24
manual SDL2 library search

jpcima:
This detects SDL2 without using the FindModule shipped with the library,
which is unreliable. It's the manual search which works in all configurations.

("old way" copied back from libOPNMIDI)
see https://github.com/Wohlstand/libADLMIDI/issues/196#issuecomment-457079484


CMakeLists.txt
```diff
@@ -233,6 +233,29 @@ else()
     add_library(ADLMIDI ALIAS ADLMIDI_static)
 endif()
 
+if(WITH_MIDIPLAY OR WITH_ADLMIDI2)
+    find_library(SDL2_LIBRARY SDL2 REQUIRED)
+    message("Found ${SDL2_LIBRARY}")
```
Flamefire: I'd rather see this as whole code as a find module. You could use ours: https://github.com/Return-To-The-Roots/s25client/blob/master/cmake/Modules/FindSDL2.cmake

CMakeLists.txt
```diff
@@ -233,6 +233,29 @@ else()
     add_library(ADLMIDI ALIAS ADLMIDI_static)
 endif()
 
+if(WITH_MIDIPLAY OR WITH_ADLMIDI2)
+    find_library(SDL2_LIBRARY SDL2 REQUIRED)
```
Flamefire: REQUIRED does not exist here

CMakeLists.txt
```diff
@@ -233,6 +233,29 @@ else()
...
+    message("Found ${SDL2_LIBRARY}")
+
+    set(SDL2_NEEDED_LIBRARIES)
```
Flamefire: Why not use standard names? (no `NEEDED` infix)

CMakeLists.txt
```diff
@@ -233,6 +233,29 @@ else()
...
+    if(WIN32)
+        if(MSVC)
+            list(APPEND SDL2_NEEDED_LIBRARIES "${SDL2_LIBRARY}")
+        else()
+            list(APPEND SDL2_NEEDED_LIBRARIES "${SDL2_LIBRARY}" pthread)
```
Flamefire: Rather use something like `find_package(Threads)` and the results of it

CMakeLists.txt
```diff
@@ -233,6 +233,29 @@ else()
...
+        endif()
+    else()
+        list(APPEND SDL2_NEEDED_LIBRARIES "${SDL2_LIBRARY}" pthread)
+        if(NOT HAIKU AND NOT OPENBSD_LOCALBASE)
+            list(APPEND SDL2_NEEDED_LIBRARIES dl)
+        endif()
+        list(APPEND SDL2_NEEDED_LIBRARIES m stdc++)
```
Flamefire: please don't link to stdc++ explicitely

CMakeLists.txt
```diff
@@ -233,6 +233,29 @@ else()
...
+    endif()
+    if(OPENBSD_LOCALBASE)
+        list(REMOVE_ITEM SDL2_NEEDED_LIBRARIES dl)
```
Flamefire: Not required as never added

jpcima:
I rewrote the whole PR @Flamefire thanks to your advice, do you think it's fine?

Flamefire:
Problem now is that you duplicate the conditions used in the subfolders. They might easily become out-of-sync. I'd suggest either:
- Duplicate the target check at the point of usage. This is kinda fine as one sees why things are done
- Create a very simple find module containing basically your code for ADLMIDI_SDL2 and `find_package` that.
- Create a wrapper function like `function(find_sdl2_targets) find_package(SDL2 ${ARGV}) <your code>` and include+call that: `find_sdl2_targets(REQUIRED) target_link_libraries(... SDL2::SDL2)`

jpcima:
Ok to merge this @Wohlstand ?
It's an annoyance to work from the master branch, because of the cmake failure.
(the case when SDL2 was built with cmake, if I understand right)

Wohlstand:
Looks much better! Yeah, I think, the same set of works are must go into libOPNMIDI's build too

PR #206, opened 2019-02-23
opal: enable use of soft panning

jpcima:
+patch Opal emulator found here
https://gist.github.com/jpcima/d4f6b99c31f3cd772a373887f2d3f318


Wohlstand:
I think, you can include the patch file also for convenience, or specify the link in README

PR #221, opened 2019-06-02
Update emulators enabled check

mmontag:
Add all emulators to the preprocessor directive that ensures at least one is enabled.

This was broken when I tried to build with only `OPAL` or `JAVA` emulator enabled.

### DIFF
--- a/src/adlmidi_opl3.cpp
+++ b/src/adlmidi_opl3.cpp
@@ -29,7 +29,10 @@
 #ifdef ADLMIDI_HW_OPL
 static const unsigned OPLBase = 0x388;
 #else
-#   if defined(ADLMIDI_DISABLE_NUKED_EMULATOR) && defined(ADLMIDI_DISABLE_DOSBOX_EMULATOR)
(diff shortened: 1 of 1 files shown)


Wohlstand:
Wops! 
Thanks for a catch! :fox_face: :+1:

mmontag:
btw, thanks for the awesome work on this project. 
I am compiling with Emscripten. With the minimal options, libADLMIDI only increased my wasm bundle by under 50 kb.  The library is super well designed and flexible.

Wohlstand:
I had to build it for Emscripten by myself also, and yeah, my little tip for you: for the work I recommend you to use the DosBox emulator which is the fastest and most accurate between fast emulators (Speaking about accuracy, the Nuked OPL3 is most accurate at all, however, on Emscripten it works choppy). :fox_face: :wink:

mmontag:
Yep, I had the same experience, and I'm now using the DosBox emulator (just had to tweak some things in order to share it between game-music-emu and libadlmidi 🙂  )

PR #222, opened 2019-06-04
WIP MIDIplay v2

jpcima:
**:warning: Not ready for merge**

Beginning of "v2" work for the MIDI player

For now, code is kept mostly as usual, there will be more rearranging.
One of the things is to split the huge structure.

It is separated in distinct source files for each logical part, if you agree with it.
Files are tagged with an `OPLMIDI` identifier.

The term of "chip voices" is adopted in file naming, does not exist yet in source.
It's the intent to be more generic, and the idea to be associated to chip channels in `1..*` way.
It's the place for a kind of polymorphic structure which can host a voice of any kind.
As such, it's intended to have the definition for "FM voice", "rhythm voice", "PSG", "PCM"  etc.

The voices will be able to be type-switched for dynamic 2-op/4-op conversion.
A tagged union will be appropriate for implementing.

Other things:
- copy constructors, as when resizing channels when adding devices/chips: eliminate
- nested structures remaining: move to upper level, rename as needed
- add generic pool of memory; it will ease implementing a buffering of note-offs, and fix in that way the "Dance of sugar-plum fairy" that Joel has noted, but **without** counting on single-time note keying.
- I want also to setup the channel<->voice Location by a better linked model which accessible by both directions

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -217,6 +217,8 @@ set(libADLMIDI_SOURCES
     ${libADLMIDI_SOURCE_DIR}/src/adlmidi_midiplay.cpp
     ${libADLMIDI_SOURCE_DIR}/src/adlmidi_opl3.cpp
     ${libADLMIDI_SOURCE_DIR}/src/adlmidi_private.cpp
+    ${libADLMIDI_SOURCE_DIR}/src/midiplay/chip_voice.cpp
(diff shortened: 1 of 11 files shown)


Wohlstand:
Looks like this needs for some updates / reworks after my recent refactoring I was done :thinking:

Wohlstand:
Closing this as this thing is extremely outdated, but ideas can be taken for the future use.

PR #224, opened 2019-07-01
dosbox: thread-safe global initialization

jpcima:
This should ensure dosbox initialization and usage in sequential manner.

Mutex functions are not called after the first time.


Wohlstand:
macOS build was broken, I have re-run the build check...

Issue #239, opened 2020-10-21
adlmidi2: fix ADLMIDI2_VIDEO_OUT_SUPPORT for Windows


HunterZ:
I noticed that `adlmidi2` has a CMake option `ADLMIDI2_VIDEO_OUT_SUPPORT` that enables support for piping video to `ffmpeg`, but that this isn't supported on Windows because of `popen()`.

Turns out Visual C++ has a compatible implementation named `_popen()`, so I was able to get it working with a few additional minor tweaks: https://twitter.com/HunterZ0/status/1318735497951211521

I will submit a pull request for consideration.

Wohlstand:
Thank you for the fix :fox_face:

PR #240, opened 2020-10-21
Fix issue Wohlstand/libADLMIDI#239

HunterZ:
Fix issue Wohlstand/libADLMIDI#239 by using _popen() plus a couple of other minor tweaks


Wohlstand:
Thanks for the fix :fox_face: :+1:

PR #242, opened 2021-02-07
Avoid home-grown integer types

Cacodemon345:
This is intended to fix compile errors on some operating systems.

### DIFF
--- a/src/wopl/wopl_file.h
+++ b/src/wopl/wopl_file.h
@@ -32,14 +32,6 @@
 extern "C" {
 #endif
 
-#if !defined(__STDC_VERSION__) || (defined(__STDC_VERSION__) && (__STDC_VERSION__ < 199901L)) \
-  || defined(__STRICT_ANSI__) || !defined(__cplusplus)
-typedef signed char int8_t;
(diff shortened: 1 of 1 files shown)


Wohlstand:
I made that as a workaround for older compiler toolchains that do lack them. They are required on older systems where they gets missing. You better change the condition to don't declare them when use modern systems and declare when using legacy

Cacodemon345:
So we bake in the check for older systems in the CMake script? That should be possible.

Wohlstand:
I mainly do check the presence of stdc macros which usually always exist and everywhere, and when they don't exist, or the C version is smaller than required, the hand-made integers getting defined. Also, I do use the C90 and C++98 standards (to ensure theb compatibility with older compilers) while I build the library, and those integers were absent at older standards.

Wohlstand:
However, I left the check for the case a wopl library can be re-used by other projects also and they can better use this.

Cacodemon345:
I still would argue going for the CMake path; it is easier to check whether the required symbols are actually defined that way.

Cacodemon345:
I think I will do this PR at a later time.

PR #260, opened 2023-02-09
CMake: added package export; fixed includes

dbarkar:
These changes will make it much easier for external users to integrate the library. Easy as:
```
find_package(libADLMIDI CONFIG REQUIRED)
...
target_link_libraries(main PRIVATE libADLMIDI::ADLMIDI)
```

- Fixed absolute path in includes
- Added static/shared targets export
- Had to increase CMake's min version for aliasing local targets in confg.cmake, there is an ugly workaround if you don't want to do this

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -313,7 +313,7 @@ if(libADLMIDI_STATIC OR WITH_VLC_PLUGIN)
     else()
         set_target_properties(ADLMIDI_static PROPERTIES OUTPUT_NAME ADLMIDI)
     endif()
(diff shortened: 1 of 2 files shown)


Wohlstand:
You should NOT bump CMake version to newer versions. That kills the support of old platforms.

dbarkar:
Alright, fixed :)

Wohlstand:
Merged that just now!
You may want to submit similar tweaks into my second project called libOPNMIDI: https://github.com/Wohlstand/libOPNMIDI
It's very similar to libADLMIDI, except for the internality.

Wohlstand:
P.S. Sorry for long delay.

PR #261, opened 2023-02-11
CMake: respect BUILD_SHARED_LIBS setting

dbarkar:
It's a good practice to respect BUILD_SHARED_LIBS setting. 
It will be easier to integrate the library to projects and build systems.


Wohlstand:
Looks better now, I'll test a bit later (I'm a bit ill, and staying at my bed right now)

PR #262, opened 2023-03-01
fix sdl2_audio build on haiku

hfsfox:


Wohlstand:
I guess I will need to handle this as fast as possible. It's weird that this thing hangs beign draft for many years and not progressed.

PR #274, opened 2023-12-11
Added an experimental ALSA sequencer client

NY00123:
For users of Linux and other relevant environments, this can be seen as a (limited) user-level ALSA equivalent of the WinMM driver.
- The given implementation uses SDL2 for outputting audio.
- The user has to select an FM bank via a command-line argument.
- Other options are not configurable (except for cases like SDL2's environment variables), and Nuked OPL3 v1.8 is always used.

I'm not as familiar with MIDI or ALSA as others may be, so good chances are that there's at least one mistake. I originally started with a small example currently found here: http://fundamental-code.com/midi/

This program could be adapted to allow greater configurability, the way the WinMM driver and other software did (like selecting a different OPL emulator or changing the number of chips). But I'm not as familiar again. Also, in this manner, I refrained from using GPLed standalone player code of ADLMIDI as a base, in case there's such code.

It currently uses SDL2 for sound output. Code can be written to use ALSA directly (as done for MIDI) and/or other APIs; I'm just more familiar with SDL2, and also know that it'll allow the use of multiple APIs/drivers via SDL_AUDIODRIVER.

For now I've used this program on my own, at least for a few purposes. Others may experiment with this client as well.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -208,6 +208,10 @@ if(WIN32)
     option(WITH_WINMMDRV    "Build a WinMM MIDI driver" OFF)
 endif()
 
+if(UNIX)
+    option(WITH_ALSASEQ     "Build an ALSA sequencer" OFF)
+endif()
+
 if(NOT WIN32
    AND NOT VITA
    AND NOT NINTENDO_3DS
@@ -461,6 +465,10 @@ if(WIN32 AND WITH_WINMMDRV)
     add_subdirectory(utils/winmm_drv)
 endif()
 
+if(UNIX AND WITH_ALSASEQ)
+    add_subdirectory(utils/adlalsaseq)
+endif()
+
 if(libADLMIDI_STATIC)
     install(TARGETS ADLMIDI_static
             EXPORT libADLMIDIStaticTargets
@@ -562,4 +570,7 @@ message("WITH_XMI2MID             = ${WITH_XMI2MID}")
 message("EXAMPLE_SDL2_AUDIO       = ${EXAMPLE_SDL2_AUDIO}")
 if(WIN32)
     message("WITH_WINMMDRV            = ${WITH_WINMMDRV}")
+endif()
+if(UNIX)
+    message("WITH_ALSASEQ             = ${WITH_ALSASEQ}")
 endif()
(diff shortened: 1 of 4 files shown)


Wohlstand:
Interesting thing!
Ye, it's better to add more arguments, there are various options including work modes (different volume models, channel allocation modes), and ability to run with an external WOPL bank files, not just with embedded banks. Also, there is an option that builds the library with all embedded banks disabled, so only external bank files can be used in such builds.

NY00123:
> Then, I could extend it by myself after testing. But right now I'm ill, and the rest of time I staying at my bed. I'll take this as soon as I can.

First of all, get well soon!

Personally, I can make use of the sequencer client separately. libADLMIDI is structured as a library so other programs can use it, after all. I've still thought of adding the ALSA sequencer client to the repository, as it's more general purpose, like the WinMM driver, albeit expected to have a significantly smaller audience.

NY00123:
btw, this can be used with Windows software that's run via Wine, as Wine may subscribe to an ALSA sequencer client. One testcase of mine is the game DemonStar (versions preceding the one recently released on Steam).

NY00123:
Add a minor modification defining the port as SND_SEQ_PORT_TYPE_MIDI_GENERIC. It was previously defined just as SND_SEQ_PORT_TYPE_APPLICATION, which wasn't sufficient for one program. Other bits might or might not make sense, and SND_SEQ_PORT_TYPE_APPLICATION should maybe be removed.

I've actually uploaded a demonstration of this sequencer with DemonStar last week. As written earlier, this does not hold to the new edition recently released via Steam, since it doesn't use MIDI at all. But other versions (including older shareware ones) should work, as demonstrated here: https://www.youtube.com/watch?v=KeTPOf9S_ks

As written earlier, I haven't really added a lot of user-accessible configuration. There might be ways in which the output differs from a single genuine OPL3 chip, like Nuked OPL3's optional extended stereo support accessible from EDuke32-based ports.

Ignoring the changes in output related to Apogee's EMIDI format, this technically allows stereo FM output for DOS executables by using the Apogee Sound System without modifying them, in case it's identified as a General MIDI device for the game. There was originally limited stereo FM support only, disabled in compile-time as of early February 1995, according to the Apogee Sound System's change log.

Wohlstand:
I started to take this thing into my `wip-alsa-seq` branch to finish it finally. I'll close ths PR since I going to finish it by myself.

NY00123:
Nice, thanks for having a look and further expanding on the idea!

PR #298, opened 2025-09-27
Added full panning support for ESFMu

kode54:
This was simple enough to implement.


Wohlstand:
Tested just now, works fine!

Wohlstand:
By the way, I want later implement the toggler so, if full-pan disabled, let output to don't produce the quiter volume. Or, just add a constant factor to compensate the centre's volume.

PR #303, opened 2025-12-19
Add PS2 support.

Wolf3s:


Wolf3s:
> Thanks! Would you like to also send the similar change to my other repos? https://github.com/Wohlstand/libOPNMIDI/ and https://github.com/Wohlstand/libEDMIDI/

yes

PR #310, opened 2026-04-16
feat: add adl_barechip_* C API exposing nuked-opl3 directly

abedegno:
Adds a small C API (`adl_barechip_*`) that exposes libadlmidi's embedded nuked-opl3 emulator at the register-write level, bypassing the MIDI / bank / sequencer layers. Useful for callers who want to render their own OPL register streams rather than drive MIDI events.

## API

```c
typedef struct ADL_BareChip ADL_BareChip;

ADLMIDI_DECLSPEC ADL_BareChip* adl_barechip_new(int sample_rate_hz);
ADLMIDI_DECLSPEC void          adl_barechip_free(ADL_BareChip* chip);
ADLMIDI_DECLSPEC void          adl_barechip_write_reg(ADL_BareChip* chip, unsigned int addr, unsigned char val);
ADLMIDI_DECLSPEC void          adl_barechip_generate(ADL_BareChip* chip, int frames, short* out_interleaved_stereo);
ADLMIDI_DECLSPEC void          adl_barechip_reset(ADL_BareChip* chip, int sample_rate_hz);
```

The wrapper owns an opaque `ADL_BareChip` that hosts a `NukedOPL3` instance via your existing `OPLChipBase`. No changes to other ABI.

## Why

My motivating use case: a C# port of the Miles AIL ADLIB.ADV TVFX engine for [Ultima Underworld 1 sound effects](https://github.com/abedegno/UnderworldGodot/pull/30) (SFX in UW1 aren't MIDI patches but proprietary byte-code state machines animating OPL2 registers at 60 Hz). I needed raw register writes + PCM generation without any MIDI layer.

More generally: chiptune tools, DOS-era game engines that store SFX as register dumps (DRO, IMF, RAW), period-accurate sound-driver ports — all want this. libadlmidi already embeds the best-in-class OPL emulator; exposing it directly saves downstream projects from vendoring a second copy of nuked-opl3.

## Why it needs to be in libadlmidi (not a separate wrapper)

The internal nuked-opl3 symbols (`OPL3_Reset`, `OPL3_WriteReg`, `OPL3_GenerateStream`) aren't marked `ADLMIDI_DECLSPEC`, so they're stripped from the Windows DLL export table and Linux `.so` dynamic symbol table. They only happen to be visible on macOS because clang's dylib linker defaults to exported-all. A downstream project can't P/Invoke them cross-platform without something in the library exporting them properly.

I originally tried P/Invoking the internal symbols directly and the macOS build worked fine, but a Windows reviewer immediately hit `EntryPointNotFoundException`. The belt-and-braces `__attribute__((visibility("default")))` on the new wrapper functions also guards against any future `-fvisibility=hidden` build.

## What's built and verified

Rebuilt cross-platform binaries and confirmed exports:

| Platform | Toolchain | Exports present |
|---|---|---|
| osx-arm64 | native clang | `nm -gU` → 5 symbols |
| linux-x64 | Docker `ubuntu:22.04` + g++ | `nm -D` → 5 symbols |
| win-x64 | Docker + mingw-w64 `x86_64-w64-mingw32-g++` | `objdump -p` → 5 exports |
| win-x86 | Docker + mingw-w64 `i686-w64-mingw32-g++` | `objdump -p` → 5 exports |

All 213 pre-existing exports on the Windows builds confirmed intact — no ABI change to the main library.

## Tradeoffs

- **OPL3 only.** `ADL_BareChip` always hosts a `NukedOPL3`. If callers wanted to pick between chip implementations (dbopl, nuked-opl2, etc.), the constructor could take an `int emulator` argument. I kept it simple since my use case is nuked-opl3-specific, but happy to broaden if you'd prefer.
- **Stereo output only.** `OPL3_Generate` writes 2 int16 per frame. OPL2-only callers just duplicate L/R.
- **No thread safety.** Same contract as the underlying `OPLChipBase` — callers serialise writes externally. Matches the rest of libadlmidi.

Happy to iterate on the API shape if you have a different preference.

### DIFF
--- a/src/adlmidi_load.cpp
+++ b/src/adlmidi_load.cpp
@@ -272,6 +272,7 @@ bool MIDIplay::LoadMIDI_post()
     //opl.Reset(); // ...twice (just in case someone misprogrammed OPL3 previously)
     m_chipChannels.clear();
     m_chipChannels.resize(synth.m_numChannels);
+
     if(m_reservedChipChannels.size() < synth.m_numChips)
         m_reservedChipChannels.resize(synth.m_numChips, 0u);
 
--- a/src/adlmidi_load.cpp
+++ b/src/adlmidi_load.cpp
@@ -272,6 +272,8 @@ bool MIDIplay::LoadMIDI_post()
     //opl.Reset(); // ...twice (just in case someone misprogrammed OPL3 previously)
     m_chipChannels.clear();
     m_chipChannels.resize(synth.m_numChannels);
+    if(m_reservedChipChannels.size() < synth.m_numChips)
+        m_reservedChipChannels.resize(synth.m_numChips, 0u);
 
     if(setToOPL2)
         synth.toggleOPL3(false);
--- a/src/adlmidi_midiplay.cpp
+++ b/src/adlmidi_midiplay.cpp
@@ -187,8 +187,10 @@ void MIDIplay::partialReset()
     chipReset();
     m_chipChannels.clear();
     m_chipChannels.resize((size_t)synth.m_numChannels);
+
     if(m_reservedChipChannels.size() < synth.m_numChips)
         m_reservedChipChannels.resize(synth.m_numChips, 0u);
+
     resetMIDIDefaults();
 }
 
@@ -559,13 +561,12 @@ bool MIDIplay::realTime_NoteOn(uint8_t channel, uint8_t note, uint8_t velocity)
             // ^ Don't use the same channel for primary&secondary
 
             // Skip chip channels reserved by the user for raw OPL writes.
-            {
-                size_t chipIdx = a / NUM_OF_CHANNELS;
-                size_t localCh = a % NUM_OF_CHANNELS;
-                if(chipIdx < m_reservedChipChannels.size() &&
-                   (m_reservedChipChannels[chipIdx] >> localCh) & 1u)
-                    continue;
-            }
+            size_t chipIdx = a / NUM_OF_CHANNELS;
+            size_t localCh = a % NUM_OF_CHANNELS;
+
+            if(chipIdx < m_reservedChipChannels.size() &&
+               (m_reservedChipChannels[chipIdx] >> localCh) & 1u)
+                continue;
 
             if(is_2op || pseudo_4op)
             {
@@ -1217,24 +1218,27 @@ size_t MIDIplay::realTime_currentDevice(size_t track)
     return m_currentMidiDevice[track];
 }
 
-void MIDIplay::realTime_rawOPL(uint8_t reg, uint8_t value)
+void MIDIplay::realTime_rawOPL2(uint8_t reg, uint8_t value)
 {
     Synth &synth = *m_synth;
+
     if((reg & 0xF0) == 0xC0)
         value |= 0x30;
+
     //std::printf("OPL poke %02X, %02X\n", reg, value);
     //std::fflush(stdout);
     synth.writeReg(0, reg, value);
 }
 
-int MIDIplay::realTime_rawOPL_Chip(size_t chipId, uint16_t reg, uint8_t value)
+int MIDIplay::realTime_rawOPL3_Chip(size_t chipId, uint16_t reg, uint8_t value)
 {
     Synth &synth = *m_synth;
+
     if(chipId >= synth.m_numChips)
         return 0;
-    // Note: intentionally no 0xC0 |= 0x30 fixup here. The public API
-    // documents that the caller is driving the chip directly and is
-    // responsible for whatever bits they want in panning/connection.
+
+    // No 0xC0 |= 0x30 fixup here: the caller is assumed to be OPL3-aware and
+    // is driving panning / connection register bits explicitly.
     synth.writeReg(chipId, reg, value);
     return 1;
 }
--- a/src/adlmidi_midiplay.cpp
+++ b/src/adlmidi_midiplay.cpp
@@ -169,6 +169,11 @@ void MIDIplay::applySetup()
     m_chipChannels.clear();
     m_chipChannels.resize(synth.m_numChannels);
 
+    // Preserve caller's chip-channel reservations across chip resets;
+    // just grow the vector to match the new chip count if needed.
+    if(m_reservedChipChannels.size() < synth.m_numChips)
+        m_reservedChipChannels.resize(synth.m_numChips, 0u);
+
     // Reset the arpeggio counter
     m_arpeggioCounter = 0;
 }
@@ -182,6 +187,8 @@ void MIDIplay::partialReset()
     chipReset();
     m_chipChannels.clear();
     m_chipChannels.resize((size_t)synth.m_numChannels);
+    if(m_reservedChipChannels.size() < synth.m_numChips)
+        m_reservedChipChannels.resize(synth.m_numChips, 0u);
     resetMIDIDefaults();
 }
 
@@ -550,6 +557,15 @@ bool MIDIplay::realTime_NoteOn(uint8_t channel, uint8_t note, uint8_t velocity)
         {
             if(ccount == 1 && static_cast<int32_t>(a) == chipChannel[0]) continue;
             // ^ Don't use the same channel for primary&secondary
+
+            // Skip chip channels reserved by the user for raw OPL writes.
+            {
+                size_t chipIdx = a / NUM_OF_CHANNELS;
+                size_t localCh = a % NUM_OF_CHANNELS;
+                if(chipIdx < m_reservedChipChannels.size() &&
+                   (m_reservedChipChannels[chipIdx] >> localCh) & 1u)
+                    continue;
+            }
 
             if(is_2op || pseudo_4op)
             {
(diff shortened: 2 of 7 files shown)


Wohlstand:
Hello!

The thing seems interesting, but it would have much more potential if you:
- Allow use of any chip emulator available in the whole libADLMIDI's set. It's a serious overkill to implement this API if you only need one single chip emulator. If you really need only a single emulator, you might want to just make a shared library over Nuked OPL3 emulator by yourself, craft a kind of the wrapper over it with making it work like a proper shared library. And then, craft a compatible C# binding to use it. In a case of Nuked OPL3, this might need a wrapper that creates the structure's pointer (make it opaque for this API), so it's pretty simple API to bind into anything will be available. And you would like to make (or use already existing call) to capture the chip's PCM output to forward it to the audio output interface to mix with other stuff or save to the file, or anything also that you want.
- It's possible to make it used when initialised as a real-time MIDI synthesizer and just adding one single `adl_rt_rawOplCommand(context, chipNum, reg, val);` call that will allow direct write of OPL2/OPL3 commands to the chip bypassing any of MIDI drivers. The thing internally is already done for needs of KSM and IMF file formats playback, they just use internally raw OPL commands sending without the use of the MIDI driver.

And so, just, libADLMIDI is an explicit MIDI driver over OPL2/OPL3 chips that works in similar manner as DMX, Apogee Sound System, AIL, etc. It handles MIDI commands, and interprets them to the real hardware or softwarely emulated chip. It would be cool if there are some cases do really require sending of raw OPL2/OPL3 register commands with bypassing of the MIDI driver, like, playing an SFX in Wolfenstein's manner, or something also. Then it's reasonable to just make one single function to send raw OPL3 commands with bypassing the MIDI driver (using the same internal method as it's used to play IMF/KSM music internally).

And again, if you only need one single emulator, then it's really purposable to build an own C# wrapper over it. If your problem that Nuke's code doesn't export anything, then craft a wrapper module that will have exported API functions that will allow you to initialise, destroy, send register commands, perform cycle, or generate PCM output, etc. etc. etc.

About thread safety you may don't worry, the library itself is not thread safe per instance. If you make an instance of the library, you are always required to use its functions with the mutex or only in the same thread. However, if you making multiple different instances, they can work safely in different threads.

Wohlstand:
Also! For the needs of SFX, it's also reasonable to implement a kind of the chip channel's reserving so MIDI driver will never allocate any data on these channels so they can be used for the SFX safely using raw OPL2/OPL3 registers.

Wohlstand:
And ye, speaking about XMI and HMI/HMP, I am working on implementing the full-featured ability to play multiple parallel MIDI files using the same driver instance that will also work on the real hardware as well as it will be more aesthetic on modern, and optimal so it's only one instance is used and no overkill of multiple instances when it's not needed (tasks like recording into WAV of multiple different songs). And I attempting to implement the rest of functionality that AIL supported so the thing will be fully suitable for games like System Shock that use the full power of the MIDI functionality.

Wohlstand (changes requested):
My final summary: Redo this. The idea has a potential, but making just singe emulator exported is an overkill for libADLMIDI's purposes. You can implement the exporting of the internal raw OPL3 commands sending, and the chip channel reserving API to make MIDI driver to not ruin your fun of raw OPL2/OPL3 writing. I wrote everything in my previous comments.

The thing can work just using the single instance of the library, and it will be suitable to use on real hardware as well.

Wohlstand:
There are different emulators provided to play the sound with different CPU load and with the different sounding accuracy. There are also OPL2 emulators included, and the extra heavy LLE emulators that implements super-accurate sounding, but they are too heavy for realtime playback even on modern computers. They will work in real time on really powerful processors, and mostly as a single chip, multichip it plays choppy. And the default output that libADLMIDI implements, is pretty flexible so you can request any format that you want for needs of your software (floats as well as integers of any size). And of course, the resampler works too, but if you build with the Zita Resampler linked, the quality of resampled data supposed to be better.

src/adlmidi_load.cpp
```diff
@@ -272,6 +272,8 @@ bool MIDIplay::LoadMIDI_post()
     //opl.Reset(); // ...twice (just in case someone misprogrammed OPL3 previously)
     m_chipChannels.clear();
     m_chipChannels.resize(synth.m_numChannels);
+    if(m_reservedChipChannels.size() < synth.m_numChips)
```
Wohlstand, line 276: Please make sure the empty line between if() block begin, and in other places where such code is used. I going soon to refactor everything for that. Just, this is for avoiding the clutter of code blocks and have it look a bit nicer and cleaner for eyes.

src/adlmidi_midiplay.cpp
```diff
@@ -182,6 +187,8 @@ void MIDIplay::partialReset()
     chipReset();
     m_chipChannels.clear();
     m_chipChannels.resize((size_t)synth.m_numChannels);
+    if(m_reservedChipChannels.size() < synth.m_numChips)
+        m_reservedChipChannels.resize(synth.m_numChips, 0u);
     resetMIDIDefaults();
```
Wohlstand, line 194: Same as above, any if blocks must have empty lines around.

src/adlmidi_midiplay.cpp
```diff
@@ -551,6 +558,15 @@ bool MIDIplay::realTime_NoteOn(uint8_t channel, uint8_t note, uint8_t velocity)
...
             // ^ Don't use the same channel for primary&secondary
 
+            // Skip chip channels reserved by the user for raw OPL writes.
+            {
+                size_t chipIdx = a / NUM_OF_CHANNELS;
+                size_t localCh = a % NUM_OF_CHANNELS;
+                if(chipIdx < m_reservedChipChannels.size() &&
+                   (m_reservedChipChannels[chipIdx] >> localCh) & 1u)
+                    continue;
+            }
```
Wohlstand: You don't need to put this into sub-block as it's no shadowing or name collision. And there are primitive types that doesn't need a destructor called quickly.

src/adlmidi_midiplay.cpp
```diff
@@ -1211,6 +1227,36 @@ void MIDIplay::realTime_rawOPL(uint8_t reg, uint8_t value)
...
 }
 
+int MIDIplay::realTime_rawOPL_Chip(size_t chipId, uint16_t reg, uint8_t value)
+{
+    Synth &synth = *m_synth;
+    if(chipId >= synth.m_numChips)
+        return 0;
+    // Note: intentionally no 0xC0 |= 0x30 fixup here. The public API
+    // documents that the caller is driving the chip directly and is
+    // responsible for whatever bits they want in panning/connection.
```
Wohlstand: The requirement of 0xC0 |= 0x30 is done for the code that expects OPL2 chip, but actually the OPL3 chip is ran in OPL3 mode. That enables the stereo, and if not specify 0x30, the output will be silent. But, for this case, the 0x30 is supposed to be added on a client side, just when they send the code for OPL2 chip while OPL3 is active, they should add 0x30 bits to make output being not silent.

src/adlmidi_midiplay.hpp
```diff
@@ -888,12 +893,44 @@ class MIDIplay
     size_t realTime_currentDevice(size_t track);
 
     /**
-     * @brief Send raw OPL chip command
+     * @brief Send raw OPL chip command (applies IMF-style connection fixup and
+     *        writes to chip 0). Used internally by IMF/KLM sequencer playback.
```
Wohlstand: I'll clarify: this command is supposed to receive the OPL2 code that doesn't expect the OPL3 chip with enabled extended mode, and that means if don't apply the 0x30 bits, the output will be just silent. And, this thing additionally is single-chip only as it writes only first chip while new-added function can write to any chip.

I think, it's possible to rename this function into "rawOPL2" with explanation that this function receives code for OPL2 chip. And ye, it can only 8 bit register addresses, so, that's why it's for OPL2 only. OPL3 has 16-bit register address (with a sub-port number added). And the new-added function can be renamed into "rawOPL3".

src/adlmidi_midiplay.hpp
```diff
@@ -888,12 +893,44 @@ class MIDIplay
...
+
+    /**
+     * @brief Reserve chip channels so the note allocator skips them.
+     *        The caller can then safely drive those channels via raw OPL
+     *        writes (adl_rt_rawOplCommand) without MIDI stealing their voices.
+     * @param chipId       Zero-based chip index [0..m_synth->m_numChips-1]
+     * @param channelMask  Bitmask of chip channel indices within the chip
+     *                     (bit 0 = channel 0, bit 17 = channel 17 for OPL3).
+     *                     Channel indices are per-chip, 0..22 (melodic 0..17,
+     *                     rhythm base 18..22).
```
Wohlstand, line 939: It's also possible to reserve 4-op channels just by reserving primary block (one 4-op channel uses two 2-op channels, one is primary that is actually used for control, and another secondary that is passive and contains only settings extension).

Wohlstand (changes requested):
This looks much better! Well done!

I do have minor notices to fix, and then I can finally accept the thing. Also, please edit the README.md to document your updates that you done in the changelog. Anyway, this is optionally, I'll document that by myself in any way.

src/adlmidi_load.cpp
abedegno: Added the empty line in 5cba546 (around the `if(m_reservedChipChannels.size() < synth.m_numChips)` block). Thanks for the style guidance.

src/adlmidi_midiplay.cpp
abedegno: Done in 5cba546 - empty lines around both `if(m_reservedChipChannels...)` blocks in `partialReset()` and `resetMIDI()`.

src/adlmidi_midiplay.cpp
abedegno: Unwrapped the sub-block in 5cba546 - now a direct `if ... continue;` with no extra scope around it.

src/adlmidi_midiplay.cpp
abedegno: Clarified the `realTime_rawOPL2` docstring in 5cba546 explaining that the 0x30 fixup exists because libadlmidi runs OPL3 in extended stereo mode; OPL2-only callers using `rawOPL3` must apply the 0x30 bits client-side if they want audible output.

src/adlmidi_midiplay.hpp
abedegno: Renamed in 5cba546 - the original `realTime_rawOPL` is now `realTime_rawOPL2` (OPL2-only, writes chip 0 with the 0x30 fixup), and the new multi-chip public entry is `realTime_rawOPL3_Chip` / `adl_rt_rawOPL3`.

src/adlmidi_midiplay.hpp
abedegno: Documented in 5cba546 on `reserveChipChannels` - noted that reserving the primary 4-op master (channels 0..5) is sufficient because the passive secondary sibling is only reached via its primary.

Wohlstand (approved):
For now, it's all fine! Other code style cases I'll fix just by myself to not overload you by extra buerocracy. Have a good time of the day!

Wohlstand:
And ye, since now, as project has the channel reseving API, now it's possible to implement the MIDI2IMF using libADLMIDI's logic: just save register commands generated by MIDI driver into the output IMF file, and that will work as well. I already have the MIDI2VGM thing at libOPNMIDI project (and I going to apply the similar one to here, but the problem that OPL chips aren't supported by libGME that is primarily used to play VGM files, and I going to implement that very soon).

Just, the primary requirement for IMF files they should don't use channel 0 as it's reserved for sound effects in games like Wolfenstein.

PR #315, opened 2026-08-25
License fixes

dbarkar:
While packaging 1.6.2 for vcpkg I went through every source file in the tree and compared what the headers say with what the repository declares. A few things do not line up.

### Two license texts were not present in the tree in any form

* `src/structures/LICENSE` — the header of `pl_list.hpp` says *"See accompanying file LICENSE or copy at http://www.boost.org/LICENSE_1_0.txt"*, and the accompanying file was never added. Text taken from boost.org.
* `src/wopl/LICENSE.txt` — the WOPL module is MIT, but unlike `src/midiseq/` and `src/models/` it had no LICENSE file of its own. Copied verbatim from the header of `wopl_file.h`.

### One directory mixes licenses

`src/midiseq/` holds MIT code (the existing `LICENSE.txt`) next to the XMIDI converter (`impl/cvt_xmi2mid.hpp`, LGPL v2+, Ryan Nunn / Bret Curtis / WildMIDI Developers). Added `LICENSE.LGPL-2.1.txt` as a second text rather than a replacement, so nothing that is already correct changes meaning; it applies through that file's "or (at your option) any later version" clause.

### README

The License section said that all Nuked OPL3 emulators are LGPL v2.1+ (`src/chips/nuked/nukedopl2.{c,h}`, Nuked OPL2 Lite, is GPL v2+) and that the MIDI sequencer is MIT (the XMIDI converter is LGPL v2+). It did not mention BSD 3-Clause, the Boost Software License, MAME OPL2, JavaOPL3, YMF262-LLE, YM3812-LLE, Opal, the OPL2/OPL3 models library or the bank notices at all. Rewritten to match the tree.

### The chipset module comes as a separate commit

`src/chips/dosbox/`, `mame/`, `java/`, `ym3812_lle/` and `ymfm/` have no license file either, and `src/chips/nuked/` ships GPL v2+ code (`nukedopl2.{c,h}`, Nuked OPL2 Lite) next to a `LICENSE` that is the LGPL v2.1 text. Since `src/chips/` is an rsync mirror of [OPL_ChipSet](https://github.com/Wohlstand/OPL_ChipSet), that half is proposed there in a companion pull request https://github.com/Wohlstand/OPL_ChipSet/pull/1 , and the first commit here is nothing but the result of running `sync_from.sh` against it.

It is kept as its own commit so you can drop it and run the sync yourself. It also carries two lines the mirror was behind on — `src/chips/chipset.cmake` and `src/chips/chipset.pri` now list `dosbox_opl2.cpp`.

`src/chips/README.md` gets the same license-list additions as the OPL_ChipSet README. Note it lists VPC OPL3 while the top-level README above does not: the chipset module can build it behind `OPL_CHIPSET_ENABLE_VPC`, libADLMIDI cannot build it at all.


Wohlstand:
Hello!
What about the midiseq, I have a campaing to eliminate these leftovers from WildMIDI like MUS2MID and XMI2MID to make the MIDISeq being fully MIT-licensed. I already remade the MUS reader into the native thing without conversion into the MIDI before reading. The same I going to do with the XMI thingies. So ye.
Thank you very much for clarifing licenses!

Wohlstand:
What about Nuked OPL2 Lite, seems it was licensed under GPLv2 by mistake, I already asked @nukeykt (the author) a question to clarify this.

PR #320, opened 2026-08-31
Corrected the license files in the repository root

dbarkar:
This file was setting the repository's label. GitHub picks license files by name, and the
interior dot in `LICENSE.GPL-3.txt` and `LICENSE.LGPL-2.1.txt` puts both outside
every pattern its detector accepts, so they score zero and are discarded before
their contents are read. The root `LICENSE` was the only candidate, which is why
the API reports `"spdx_id": "LGPL-3.0"` for the whole project today.

That label is wrong in the dangerous direction. It tells a reader that linking from
proprietary code is permitted, and it is not: the public header and the core
translation units are GPL v3+, so anything that links the library takes on
GPL-3.0-or-later. Somebody who trusts the badge and ships a closed product on it
ends up in violation.

A repository label describes the combined work, and for libADLMIDI that is
GPL-3.0-or-later. LGPL 2.1 covers components only - the chip emulators and the
interfaces to them - and the components converge upward rather than pulling the
aggregate down: the GPL-2.0-or-later emulators upgrade through their own "or any
later version", and the LGPL-2.1-or-later parts upgrade both through LGPL 2.1 §3,
which expressly permits opting into a newer ordinary GPL, and through their own
or-later clause.

What settles it is the direction of the error. An LGPL-2.1 label understates the
restrictions and invites a violation. A GPL-3 label overstates them for some
components, but never grants a right that does not exist. That is the only
direction in which it is safe to be wrong.

A label is also not a relicensing. The LGPL 2.1 parts stay LGPL 2.1, exactly as the
README's License section describes them. The badge is a coarse marker for tooling;
the per-component detail belongs in the README, where it already is.

The resulting names also read sensibly: `LICENSE.txt` is the license of the work as
a whole, `LICENSE.LGPL-2.1.txt` is a component license.


dbarkar:
Also small fix for readme

Wohlstand:
Thanks for clarifications!

Wohlstand:
For the future, I have a plan to get rid of leftovers of Biswqit's code parts (since libADLMIDI is built on Bisqwit's base, the GPLv3+ license is set on the whole project for now) and replacing them with my own coded as I going to allow the library itself to be licensed with more permissive license, and it will directly depend on what emulators are enabled (it's will be possible to have LGPL if disable emulators are GPL licensed, or MIT if disable emulators are LGPL-licensed and keep public domain and BSD licensed one).

Initially I wanted to request LGPL from Bisqwit, but he challenged me, I made for him the thing that he requested, but then he requested to split it to a sequence of small pieces. And even worse, he responts extremely rare (I don't get WHY he makes such buerocratic hell for years?). And it's more reasonable to completely rewrite internals so they no longer to be based on Bisqwit's code. And ye, I think, I will completely drop the support for auto-arpeggio feature, as it works mostly harmful than assisting. It looks like a good experiment, but in a production it's a mess, many people complained me how it sucks... For now, I made it being disabled by default at least.

PR #324, opened 2026-09-16
Fix two cases of undefined behaviour that UBSan reports

yumasansansan:
Two places in the emulator cores rely on undefined behaviour.
UndefinedBehaviorSanitizer stops the program at both, which makes a
sanitizer build of anything that uses libADLMIDI impossible to run.

**DBOPL** (`src/chips/dosbox/dbopl.cpp`): `InitTables()` builds the tables of
the channels' and the operators' offsets from the start of the chip by taking
the address of a member through a null pointer.

```
src/chips/dosbox/dbopl.cpp:1621:47: runtime error: member access within null
pointer of type 'DBOPL::Chip'
```

`offsetof` gives the same numbers and is defined for these types: `Chip`,
`Channel` and `Operator` have no base classes and no virtual functions, and
all of their data members sit in one public section, so they are
standard-layout.

**JavaOPL3** (`src/chips/java/JavaOPL3.hpp`): the attack rates 0 to 3 have an
infinite period in the table of the YMF278B manual's values (`MUGEN`), and
`setActualAttackRate()` converts that infinity to `int`.

```
src/chips/java/JavaOPL3.hpp:1411:35: runtime error: inf is outside the range
of representable values of type 'int'
```

The numbers it works out do not matter there. The increment is the reciprocal
of the period, which is zero for an infinite one, and it multiplies the
difference of the two periods in samples, so the minimum of the attack comes
out the same whatever those two are. An infinite period is answered directly
now, with the values the rest of the function would have reached.

Neither changes the sound. Rendering the same sequence before and after gives
the same bytes: DOSBox against a hash committed before the change, Java OPL3
compared directly.

Found while building [ADLplug-Next](https://github.com/yumasansansan/ADLplug-Next)
with the address and undefined sanitizers.


Wohlstand:
... what the heck robot intruded pull-requests to my repos? I don't remeber I even enabled this.

Wohlstand:
I reviewed the thing, and looks interesting. However, I do have one small question.

src/chips/ym3812_lle/nopl2.c
```diff
@@ -103,7 +103,7 @@ static void nopl2_cycle(nopl2_t *chip)
                 if (shift)
                 {
                     shift--;
-                    chip->sample = mant << shift;
+                    chip->sample = mant * (1 << shift);
```
Wohlstand, line 106: I am curious what did you attempted to solve by this? Does this solves some kind of the sound distortion?

Wohlstand:
P.S. Did you based your ADLPlug fork on [my repo](https://github.com/Wohlstand/ADLplug/)? I made several fixes on it and upgraded support for libADLMIDI/libOPNMIDI thingies. If you did not yet imported these fixes, please import. And ye, I do support my fork mostly for minor maintenance since JPCima was suddenly gone into nowhere. And I am appriciated that you decided to take the effort to success this thing. :) I'll try to give it to my buddy at the chat for his tests.

yumasansansan:
> ... what the heck robot intruded pull-requests to my repos? I don't remeber I even enabled this.

It might be because of my "Automatic Copilot code review" feature. I've disabled it now, but it's possible that other features are also interfering unexpectedly.

Disclaimer: Since I'm not a native English speaker — I am Japanese, I use LLMs and translation websites for writing documents and replying to messages. Although I proofread my work, there may still be errors. I appreciate your understanding.

> I am curious what did you attempted to solve by this? Does this solves some kind of the sound distortion?

No, it does not change the sound — that is rather the point. `mant` is `(chip->shifter & 0x3ff) - 512`, so it is often negative, and in C a left shift of a negative value is undefined behaviour. UBSan stops the program there, which means libADLMIDI cannot be run under a sanitizer:
```
src/chips/ym3812_lle/nopl2.c:106:41: runtime error: left shift of negative value -512
```
The multiplication is the same arithmetic, and Clang emits the same code for it: I compared the disassembly of the object file with and without the change and it is identical, byte for byte. So it costs nothing at run time — it only stops the compiler from being allowed to assume the shift never happens.

For what it is worth, the low-level cores do this in many more places: the gate-level latches are ints that shift bits along until they pass the sign. That is far too many places to patch, so my build compiles `ym3812_lle`, `ymf262_lle` and libOPNMIDI's `nuked_lle` with `-fwrapv` instead. This one was a single expression, so it seemed worth fixing at the source.

> P.S. Did you based your ADLPlug fork on my repo? I made several fixes on it and upgraded support for libADLMIDI/libOPNMIDI thingies. If you did not yet imported these fixes, please import. And ye, I do support my fork mostly for minor maintenance since JPCima was suddenly gone into nowhere. And I am appriciated that you decided to take the effort to success this thing. :) I'll try to give it to my buddy at the chat for his tests.

ADLplug-Next branches from jpcima/ADLplug at `a488abe`, the head of his master, so no — I had not picked your fork up. Thank you for pointing it out; I went through the 22 commits you have on top of that:

- The submodule bumps and the CMake/CI work I have already gone past — ADLplug-Next is on JUCE 9 with its own CMake and a CI that builds and packages for Windows, Ubuntu and macOS.
- "Don't show VGM dumper in emulators": done here too (`USE_VGM_FILE_DUMPER OFF`).
- "Update wopl_file.h" does not apply any more — there is no separate wopl copy, it uses libADLMIDI's.
- **"Backend: Added Channel allocation mode option" is a real gap.** I do not have it, and I will bring it in, with credit to you.

And thank you, that means a lot coming from you. If your friend would like something to try, there are nightly builds for the three systems in the Releases of https://github.com/yumasansansan/ADLplug-Next — it is still moving quickly, so reports are very welcome.

Wohlstand:
> Disclaimer: Since I'm not a native English speaker — I am Japanese, I use LLMs and translation websites for writing documents and replying to messages. Although I proofread my work, there may still be errors. I appreciate your understanding.

心配しないでください、私は自分で英語を話すことはありません。 私の母国語はロシア語です。 :)

> src/chips/ym3812_lle/nopl2.c:106:41: runtime error: left shift of negative value -512

I see, ye, usually this works normally on desktops, but on SOME platforms it might seriously mess. And ye, that emulator is always disabled by default with other "-LLE" since they are too heavy for ordinary computers and no way they can work on mobile devices. They mostly suitable for renrering into WAV without realtime playthrough since these emulators are too slow for the rest of cases.

And ye, I going to merge this. Sorry for a delay, I was too busy on my main work.

PR #325, opened 2026-09-16
WOPL: a version of 0 is not a version

yumasansansan:
The versions of WOPL are 1 to 3, but `WOPL_LoadBankFromMem()` and `WOPL_LoadInstFromMem()` turn away only a version newer than they know (`WOPL_ERR_NEWER_VERSION`), so they accept a version of 0. Saving uses 0 to mean "the latest version", so such a file does not survive being saved in its own version and loaded back: a bank loaded as version 0 is saved and comes back as version 3.

A fuzzer found this within seconds, with a bank of 152 bytes: the magic `WOPL3-BANK\0` followed by zeros.

This makes both loaders turn a version of 0 away, with a new error code appended to the end of `WOPL_ErrorCodes`, `WOPL_ERR_INVALID_VERSION`, so no existing value changes. `adlmidi_load.cpp` reports it as "Custom bank: Invalid version!" instead of "Unknown error!". Files written by any tool are not affected, since none writes a version of 0.

Found while fuzzing the bank and instrument loading of [ADLplug-Next](https://github.com/yumasansansan/ADLplug-Next) with libFuzzer and the address and undefined sanitizers. libOPNMIDI's WOPN loaders have the same gap, which I am sending there as well.


Wohlstand:
A good catch! And ye, this can be cherry picked to my [OPL3 Bank Editor too](https://github.com/Wohlstand/OPL3BankEditor/tree/master/src/FileFormats/wopl).

PR #329, opened 2026-09-19
Nuked OPL3: the fast path silences a slot as the other path does

yumasansansan:
The fast waveform generator has no clamp on the level it shifts the exponent table by, and the level can pass 0x1fff:

    src/chips/nuked/nukedopl3.c:913:38: runtime error: shift exponent 32 is too large for 32-bit type 'int'

`OPL3_EnvelopeCalcExp()`, a few hundred lines up in the same file, holds the level at 0x1fff for exactly this reason: the shift is then 31 at the most, and a slot attenuated past the end of the table comes out silent. `OPL3_SlotGenerate()` under `OPL_FAST_WAVEGEN` — which is the default — works the level out for itself and does not hold it, so a level of 0x2000 or more shifts by 32 or more.

That is undefined, and on x86 it is the shift by nothing at all: the slot comes out at full level where it should be silent. So this is audible as well as undefined, a burst instead of a silence. The clamp added here is the one from `OPL3_EnvelopeCalcExp()`, in the same words; every level up to 0x1fff is untouched.

A fuzz target of [ADLplug-Next](https://github.com/yumasansansan/ADLplug-Next) found it, one that plays MIDI through the real-time API with the chip settings a saved project can hold. Twelve bytes were enough: thirty-five four-operator channels and a drum note of 32. The daily run and a manual run both stopped there.

Small and separate, as #326 said it would be: this one touches `src/chips/nuked/nukedopl3.c` and nothing else.


yumasansansan:
I have added a second commit to this branch, for the other undefined shift in the same function.

`OPL3_SlotGenerate()` takes the phase through `slot->phaseshift`, and for waveform 7 that shift is 32 -- "no shift (work by mod 32), but flag for non-sin wave", as the line that sets it says. The phase is a `uint16_t`, so the shift is by more than the width of the value:

      src/chips/nuked/nukedopl3.c:904:11: runtime error: shift exponent 32 is too large for 16-bit type 'uint16_t'

Waveform 6, whose shift is 16, is undefined for a second reason: the phase is promoted to a signed `int`, and a phase with its top bit set shifts into the sign. The commit says `mod 32` where the comment already meant it and takes the shift in an unsigned value wide enough for it, so that waveform 7 comes through unshifted, waveform 6 comes out as nought, and the sine waveforms shift by nought or one, exactly as they do today. Compiled with `-O2` and with `-O3`, the object file is byte for byte the one from before.

The same fuzz target found it as found the level past the table: it plays MIDI through the real-time API with the chip settings a saved project can hold, and the input that stopped it is eighteen bytes which ask for the library's own default emulator.

